PR #2, opened 2015-03-07
Remove .DS_Store file that was accidentally included.

mattwala:


mattwala:
https://gitlab.tiker.net/inducer/sumpy/merge_requests/8

Issue #11, opened 2020-08-04
Fourier-Bessel (possibly other) multipoles fail to form (produce NaNs) if there's a source coincident with a center


inducer:
x-ref: https://gitlab.tiker.net/xywei/volumential/-/issues/43

cc @xywei @isuruf

PR #12, opened 2020-08-04
Add nan fail test

xywei:
Closes https://github.com/inducer/sumpy/issues/11

cf. https://gitlab.tiker.net/xywei/volumential/-/issues/43

### DIFF
--- a/test/test_kernels.py
+++ b/test/test_kernels.py
@@ -128,7 +128,8 @@ def test_p2p(ctx_factory, exclude_self):
     False,
     True
     ])
-# Sample: test_p2e_with_source_at_center_no_nans(cl._csc, HelmholtzKernel(2), H2DMultipoleExpansion, 4, False)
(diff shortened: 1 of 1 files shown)


inducer:
Thanks!

cc @isuruf

isuruf:
`HelmholtzConformingVolumeTaylorMultipoleExpansion` works? `H2DMultipoleExpansion` will not be the default Helmholtz expansion in the near future.

xywei:
> `HelmholtzConformingVolumeTaylorMultipoleExpansion` works? `H2DMultipoleExpansion` will not be the default Helmholtz expansion in the near future.

I think that Taylor expansions are okay because they are smooth. The issue pertains only to basis functions that are singular (like Bessel functions of the second kind).

inducer:
> I think that Taylor expansions are okay because they are smooth.

Taylor multipoles aren't smooth though. (The multipole basis consists of derivatives of the kernel.)

PR #16, opened 2020-10-02
Convert PDE system to scalar PDE

isuruf:


sumpy/expansion/diff_op.py
```diff
@@ -119,12 +140,87 @@ def to_sym(self, fnames=None):
...
+def as_scalar_pde(pde, vec_idx):
+    r"""
+    Returns a scalar PDE that satisfies the `vec_idx` component
+    of a vector-valued function.
+
+    :arg pde: An instance of :class:`LinearPDESystemOperator`
+    :arg vec_idx: the index of the vector-valued function that we
+                  want as a scalar PDE
+    """
+    from six import iteritems
```
inducer, line 153: #17 drops/has dropped `six`.

sumpy/expansion/diff_op.py
```diff
@@ -157,25 +252,52 @@ def gradient(diff_op):
...
-    :arg ninput: number of input variables to the function
+    :arg ninput: number of spatial variables of the function
     :arg noutput: number of output values of function
+    :arg include_time: include time as a dimension
     """
-    mi = tuple([0]*ninput)
+    if include_time:
+        mi = tuple([0]*(ninput + 1))
+    else:
+        mi = tuple([0]*ninput)
```
inducer, line 301: Why is time being handled specially here?

inducer (commented):
LGTM mostly. Two minor wrinkles.

inducer:
Thanks for working on this!

sumpy/expansion/diff_op.py
isuruf: Issue was that when we say `laplacian(u)` it means laplacian with spatial variables. We don't want to do `d^2/dx^2 + d^2/dy^2 + d^2/dt^2`. I agree that it's not pretty, and I'm open to suggestions.

inducer: I guess just document that there's a convention that the first `dim` partial derivatives in the PDE operator refer to the spatial variables.

inducer:
LGTM, thanks!

PR #19, opened 2020-10-13
Allow multiple kernels in p2e

isuruf:
Needs https://github.com/inducer/boxtree/pull/20
Testing pytential in inducer/pytential#33

PR #23, opened 2020-11-04
Allow one FMM to have multiple source kernels

isuruf:
- [x] Depends on https://github.com/inducer/sumpy/pull/19
- [x] DO NOT MERGE until https://github.com/inducer/pytential/pull/38 is ready (hence draft)


inducer:
I'm guessing that this will need a corresponding pytential change, for p2qbxl. Could you start a test branch?

sumpy/p2e.py
```diff
@@ -57,16 +57,23 @@ def __init__(self, ctx, expansion, kernels=None,
...
+        else:
+            kernels = kernels
+            for knl in kernels:
+                assert tdr(knl) == knl
+                assert sdr(knl) == expansion.kernel
 
-        from sumpy.kernel import TargetDerivativeRemover
-        expansion = expansion.with_kernel(
-                TargetDerivativeRemover()(expansion.kernel))
+        KernelComputation.__init__(self, ctx=ctx, out_kernels=[], in_kernels=kernels,
```
inducer, line 74: Use `super()`?

sumpy/p2e.py
```diff
@@ -57,16 +57,23 @@ def __init__(self, ctx, expansion, kernels=None,
           number of strength arrays that need to be passed.
           Default: all kernels use the same strength.
         """
-        if kernels is None:
-            kernels = [expansion.kernel]
+        from sumpy.kernel import TargetDerivativeRemover, SourceDerivativeRemover
+        tdr = TargetDerivativeRemover()
+        sdr = SourceDerivativeRemover()
+        expansion = expansion.with_kernel(tdr(expansion.kernel))
+        expansion = expansion.with_kernel(sdr(expansion.kernel))
```
inducer, line 64: Expansions should only be constructed with the base kernel.

sumpy/p2e.py
```diff
@@ -57,16 +57,23 @@ def __init__(self, ctx, expansion, kernels=None,
...
-        KernelComputation.__init__(self, ctx=ctx, kernels=kernels,
-                strength_usage=strength_usage, value_dtypes=None,
-                name=name, device=device)
+        if kernels is None:
+            kernels = [tdr(expansion.kernel)]
+        else:
+            kernels = kernels
+            for knl in kernels:
+                assert tdr(knl) == knl
+                assert sdr(knl) == expansion.kernel
```
inducer, line 72: What is `kernels` for? Can't we supply kernels at expansion time (rather than when calling the constructor)?

sumpy/p2p.py
```diff
@@ -52,76 +51,128 @@
 # {{{ p2p base class
 
 class P2PBase(KernelComputation, KernelCacheWrapper):
-    def __init__(self, ctx, kernels, exclude_self, strength_usage=None,
-            value_dtypes=None, name=None, device=None):
+    def __init__(self, ctx, out_kernels, exclude_self, strength_usage=None,
+            value_dtypes=None, name=None, device=None, in_kernels=None):
```
inducer, line 55: I'm not sure I get why `in_kernels` and `out_kernels` makes sense for P2P. Could you explain?

inducer (commented):
Some thoughts below.

sumpy/p2e.py
isuruf: Yes, kernels are only needed at expansion time and is unnecessary in the constructor, but the way that caching is set up, we need them in the constructor. Overhauling it is a good thing but we would need some unrelated work here in sumpy and in pytential

sumpy/p2p.py
isuruf: Let `ds` be some source derivative and `dt` be some target derivative.
To calculate  `H1 = ds(G) + G` and `H2 = dt(ds(G) + G)`, we need to calculate `ds(G), G` first which are the `in_kernels` and then the `out_kernels` would be `G, dt(G)` corresponding to `H1, H2`.

sumpy/p2e.py
inducer: We'll defer that change to another PR.

sumpy/p2p.py
```diff
@@ -52,76 +51,128 @@
...
+            for i, out_knl in enumerate(self.target_kernels):
+                expr = out_knl.postprocess_at_target(
+                        in_knl.postprocess_at_source(
+                            in_knl.get_expression(dvec),
                             dvec),
                         dvec)
-                    )
-                for i, knl in enumerate(self.kernels)]
+                expr *= self.get_strength_or_not(isrc_sym, 0)
+                exprs.append(expr)
```
inducer, line 119: How come this must be treated specially? (Aka I would much prefer if this branch didn't exist.)

sumpy/p2p.py
```diff
@@ -185,25 +235,26 @@ def get_kernel(self):
...
                 nstrengths=self.strength_count,
-                nresults=len(self.kernels)),
+                nresults=len(self.target_kernels)),
             lang_version=MOST_RECENT_LANGUAGE_VERSION)
 
         loopy_knl = lp.tag_inames(loopy_knl, "idim*:unr")
 
-        for knl in self.kernels:
+        for knl in self.target_kernels:
             loopy_knl = knl.prepare_loopy_kernel(loopy_knl)
```
inducer, line 258: Is there a good reason this is only done for target kernels?

sumpy/p2p.py
```diff
@@ -264,7 +313,7 @@ def get_kernel(self):
...
```
inducer, line 317: As above: Is there a good reason this is only done for target kernels?

sumpy/p2p.py
```diff
@@ -340,7 +387,7 @@ def get_kernel(self):
         loopy_knl = lp.add_dtypes(loopy_knl,
             dict(nsources=np.int32, ntargets=np.int32))
 
-        for knl in self.kernels:
+        for knl in self.target_kernels:
             loopy_knl = knl.prepare_loopy_kernel(loopy_knl)
```
inducer, line 391: As above: Is there a good reason this is only done for target kernels?

sumpy/p2p.py
```diff
@@ -482,7 +530,7 @@ def get_kernel(self):
         loopy_knl = lp.tag_array_axes(loopy_knl, "targets", "sep,C")
         loopy_knl = lp.tag_array_axes(loopy_knl, "sources", "sep,C")
 
-        for knl in self.kernels:
+        for knl in self.target_kernels:
             loopy_knl = knl.prepare_loopy_kernel(loopy_knl)
```
inducer, line 534: As above: Is there a good reason this is only done for target kernels?

sumpy/p2p.py
```diff
@@ -472,7 +520,7 @@ def get_kernel(self):
             fixed_parameters=dict(
                 dim=self.dim,
                 nstrengths=self.strength_count,
-                nkernels=len(self.kernels)),
+                nkernels=len(self.target_kernels)),
```
inducer, line 523: The variable name `nkernels` should be changed. `ntarget_kernels`? (Not that I like this very much.) `noutputs`? (In preparation of introducing "outputs" that aren't kernels?)

(Here and elsewhere)

sumpy/qbx.py
```diff
@@ -62,45 +62,67 @@ def stringify_expn_index(i):
...
+
+        self.dim = single_valued(knl.dim for knl in self.target_kernels)
+        self.expansion = expansion
 
     def get_cache_key(self):
-        return (type(self).__name__, tuple(self.kernels),
-                tuple(self.strength_usage), tuple(self.value_dtypes))
+        return (type(self).__name__, self.expansion, tuple(self.target_kernels),
+                tuple(self.source_kernels), tuple(self.strength_usage),
+                tuple(self.value_dtypes))
```
inducer, line 99: Add device?

sumpy/qbx.py
```diff
@@ -62,45 +62,67 @@ def stringify_expn_index(i):
...
+        if source_kernels is None and target_kernels is None:
+            from warnings import warn
+            warn("A list for the second argument of the constructor is deprecated."
+                "Use a single expansion and pass source_kernels and target_kernels",
+                DeprecationWarning, stacklevel=2)
+            source_kernel = single_valued(tdr(exp.kernel) for exp in expansion)
+            base_kernel = sdr(source_kernel)
+            target_kernels = tuple(sdr(exp.kernel) for exp in expansion)
+            source_kernels = (source_kernel,)
+            expansion = expansion[0].copy(kernel=base_kernel)
```
inducer, line 87: I don't think this can work. The other expansion could be a multipole, which would compute something totally different. Fortunately, I don't think this is used. I think the only reasonable way forward is to remove this ability outright.

sumpy/qbx.py
```diff
@@ -157,7 +183,8 @@ def get_default_src_tgt_arguments(self):
                     None, shape="ntargets"),
                 lp.ValueArg("nsources", None),
                 lp.ValueArg("ntargets", None)]
-                + gather_loopy_source_arguments(self.kernels))
+                + gather_loopy_source_arguments(
+                    self.target_kernels + self.source_kernels))
```
inducer, line 187: Elsewhere, this is only called for the source kernels. Why the inconsistency?

sumpy/qbx.py
```diff
@@ -243,7 +272,7 @@ def get_kernel(self):
             lang_version=MOST_RECENT_LANGUAGE_VERSION)
 
         loopy_knl = lp.tag_inames(loopy_knl, "idim*:unr")
-        for expn in self.expansions:
+        for expn in self.target_kernels + self.source_kernels:
             loopy_knl = expn.prepare_loopy_kernel(loopy_knl)
```
inducer, line 276:
```suggestion
        for knl in self.target_kernels + self.source_kernels:
            loopy_knl = knl.prepare_loopy_kernel(loopy_knl)
```

sumpy/p2p.py
```diff
@@ -52,76 +51,128 @@
...
+        base_target_kernel = single_valued(tdr(knl) for knl in target_kernels)
+        assert base_source_kernel == base_target_kernel
+
+        KernelComputation.__init__(self, ctx=ctx, target_kernels=target_kernels,
+            source_kernels=source_kernels, strength_usage=strength_usage,
+            value_dtypes=value_dtypes, name=name, device=device)
 
         self.exclude_self = exclude_self
+        self.source_kernels = source_kernels
+        self.target_kernels = target_kernels
```
inducer, line 88: This is also set in `KernelComputation`. Why set it again here?

test/test_fmm.py
```diff
@@ -198,6 +198,95 @@ def test_sumpy_fmm(ctx_factory, knl, local_expn_class, mpole_expn_class):
     pconv_verifier()
 
 
+def test_unified_single_and_double(ctx_factory):
```
inducer, line 201: Add a docstring description?

inducer (commented):
Thanks for working on this. (And happy new year!) A few comments below. OK from my end once these are addressed.

sumpy/p2p.py
isuruf: Used noutputs

sumpy/qbx.py
isuruf: Fails with ` unsupported type for persistent hash keying: <class 'pyopencl._cl.Device'>`

sumpy/p2p.py
inducer: Got it, thx.

inducer (approved):
This looks good, thanks for working on it. I'll look over the pytential part, and if both are OK, they go in.

inducer:
Also, I made #25 to make sure the examples continue to work and merged it in here. If it fails, could you fix it? (Sorry, and thanks!)

inducer:
Hmm, that didn't take long. :)

PR #37, opened 2021-04-05
Kill kill_trivial_assignments

isuruf:
Only a handful of assignments get killed with this. For eg: 100K total -> 3 assignments killed.


inducer:
Yay!

PR #41, opened 2021-04-13
Allow offsets in P2P kernel

alexfikl:
Required by inducer/pytential#71. (turns out it was just a cache invalidation issue!)


inducer:
Thanks!

PR #43, opened 2021-04-18
Use new ci-support to upload benchmarks

isuruf:


inducer:
I think this'll need a way to be disabled. Otherwise it'll keep breaking the Github CI. (it has, on main, already)

isuruf:
Did it break other projects too?

inducer:
No, I think those were saved by not having an `.asv.conf`.

inducer:
Thanks!

PR #46, opened 2021-04-26
Add TargetPointMultiplier

isuruf:
To reproduce the results of https://doi.org/10.1016/j.jcp.2007.06.029

- [ ] Review https://github.com/inducer/pytential/pull/75 and merge at the same time

### DIFF
--- a/sumpy/e2p.py
+++ b/sumpy/e2p.py
@@ -58,14 +58,14 @@ class E2PBase(KernelCacheWrapper):
 
         from sumpy.kernel import (SourceTransformationRemover,
                 TargetTransformationRemover)
-        sdr = SourceTransformationRemover()
-        tdr = TargetTransformationRemover()
+        sxr = SourceTransformationRemover()
+        txr = TargetTransformationRemover()
         expansion = expansion.with_kernel(
-                sdr(expansion.kernel))
+                sxr(expansion.kernel))
 
-        kernels = [sdr(knl) for knl in kernels]
+        kernels = [sxr(knl) for knl in kernels]
         for knl in kernels:
-            assert tdr(knl) == expansion.kernel
+            assert txr(knl) == expansion.kernel
 
         self.ctx = ctx
         self.expansion = expansion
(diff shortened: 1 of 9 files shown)


sumpy/e2p.py
```diff
@@ -56,9 +56,10 @@ def __init__(self, ctx, expansion, kernels,
         if device is None:
             device = ctx.devices[0]
 
-        from sumpy.kernel import SourceDerivativeRemover, TargetDerivativeRemover
-        sdr = SourceDerivativeRemover()
-        tdr = TargetDerivativeRemover()
+        from sumpy.kernel import (SourceTransformationRemover,
+                TargetTransformationRemover)
+        sdr = SourceTransformationRemover()
+        tdr = TargetTransformationRemover()
```
inducer, line 62:
```suggestion
        sxr = SourceTransformationRemover()
        txr = TargetTransformationRemover()
```
* Nb: avoid shadowing `str` :)
* Analogous below.

sumpy/qbx.py
```diff
@@ -183,29 +183,30 @@ def get_kernel(self):
         raise NotImplementedError
 
     def get_optimized_kernel(self,
-            targets_is_obj_array, sources_is_obj_array, centers_is_obj_array):
+            targets_is_obj_array, sources_is_obj_array, centers_is_obj_array,
+            itgt_name="itgt"):
```
inducer, line 187: For my education: where does `itgt` need to be overridden?

inducer (commented):
LGTM, just two minor things.

sumpy/qbx.py
isuruf: It's used at https://github.com/inducer/pytential/pull/75/files#diff-4fcf38e9154a831bf71a50806d418d2ce5fbdd57cdf44c979ffc90f919e03b1dR123-R127

inducer:
```suggestion
            # Used by pytential to override the name of the loop to be
            # parallelized. In the case of QBX, that's the loop over QBX
            # targets (not global targets).
            itgt_name="itgt"):
```

inducer: FWIW, it took me about 20 minutes of puzzling to convince myself that what you're doing here is correct, and I'm not sold the rename makes things clearer... to be fair, I'm not claiming that the code was clear or beautiful to start with.

inducer:
Looks OK, thanks.

PR #51, opened 2021-05-11
Add ElasticityKernel

isuruf:


inducer:
Thanks, LGTM! Any idea why pytential is unhappy?

sumpy/kernel.py
```diff
@@ -662,44 +664,62 @@ def get_pde_as_diff_op(self):
...
 
-class StokesletKernel(ExpressionKernel):
-    init_arg_names = ("dim", "icomp", "jcomp", "viscosity_mu_name")
+class ElasticityKernel(ExpressionKernel):
+    init_arg_names = ("dim", "icomp", "jcomp", "viscosity_mu_name",
+            "poisson_ratio")
 
-    def __init__(self, dim, icomp, jcomp, viscosity_mu_name="mu"):
+    def __new__(cls, dim, icomp, jcomp, viscosity_mu="mu", poisson_ratio="nu"):
+        if poisson_ratio == "0.5":
```
inducer, line 672: Is a string really the best solution here? Could you explain?

isuruf: No, it isn't. I tried pymbolic expressions, but it fails with,
```
TypeError: unsupported type for persistent hash keying: <class 'pymbolic.primitives.Variable'>
```
for `mu`.

inducer:
Thanks!

sumpy/kernel.py
inducer: Thanks for making this happen.

PR #56, opened 2021-05-18
relax the order requirement a bit

isuruf:


test/test_misc.py
```diff
@@ -115,7 +115,7 @@ def test_pde_check_kernels(ctx_factory, knl_info, order=5):
         eoc_rec.add_data_point(h, err)
 
     print(eoc_rec)
-    assert eoc_rec.order_estimate() > order - knl_info.nderivs + 1 - 0.1
```
inducer, line 118: Can you add a bit of context? It seems to pass with the stricter requirement. Why loosen it? (I'm guessing this is for the benefit of something else that's not in yet... could we loosen it just for that?)

isuruf: It fails at https://github.com/inducer/sumpy/pull/46/checks?check_run_id=2608268658 which is an unrelated PR. The failures are random and don't always occur.

inducer: I'd at least make that loosening specific to the kernels where it's needed.

> The failures are random and don't always occur.

That's super concerning IMO. There's no randomness in `test_pde_check_kernels`. Could you take a look to see whether it's feasible to understand the source of nondeterminism?

isuruf: I haven't been able to reproduce locally (tried 100 times). In CI, it's pretty frequent. (1 in 3 times maybe)

inducer: You can SSH into CI using https://github.com/marketplace/actions/debugging-with-tmate.

isuruf: Randomness is because of https://github.com/inducer/sumpy/blob/3b874c9591e5a96775aea185aa01dd4741e64462/test/test_misc.py#L101

inducer: :facepalm: Sorry! #57

isuruf: Thanks

PR #58, opened 2021-05-22
Add a order property to LinearPDESystemOperator

isuruf:


sumpy/expansion/diff_op.py
```diff
@@ -65,6 +65,13 @@ def __eq__(self, other):
     def __hash__(self):
         return hash((self.dim, self.eqs))
 
+    @property
+    def degree(self):
```
inducer, line 69: I think "order" is the more conventional term. Would that work for you?

isuruf: Yes, works for me.

test/test_misc.py
```diff
@@ -411,8 +411,11 @@ def test_as_scalar_pde_elasticity():
     ]
 
     pde = concat(*exprs)
+    assert pde.degree == 1
```
inducer, line 414:
```suggestion
    assert pde.order == 1
```

test/test_misc.py
```diff
@@ -411,8 +411,11 @@ def test_as_scalar_pde_elasticity():
...
     for i in range(5):
-        assert as_scalar_pde(pde, i) == laplacian(laplacian(diff_op[0]))
+        scalar_pde = as_scalar_pde(pde, i)
+        assert scalar_pde == laplacian(laplacian(diff_op[0]))
+        assert scalar_pde.degree == 4
```
inducer, line 418:
```suggestion
        assert scalar_pde.order == 4
```

inducer:
Thanks!

test/test_misc.py
isuruf: Thanks. Forgot to push this change.

Issue #61, opened 2021-05-31
e2p for H2DLocalExpansion evaluates Bessel functions redundantly in test_p2e2p


inducer:
(This happens in connection with the second output, a `DirectionalSourceDerivative`.)

cc @isuruf

isuruf:
I'm not sure I understand what the issue is. Can you explain?

inducer:
Sure! `test_p2e2p` computes two "target kernels", one the plain kernel, the other a source derivative. Both involve evaluation of Bessel functions of different orders, but with the same argument. (The base kernel needs `J3` at most, the derivative up to `J5`.) They both take the same argument, but the CSE processing isn't smart enough to recognize that, leading to redundant evaluation of the Bessel function.

PR #64, opened 2021-06-02
Add new Hankel1 and BesselJ functions that work with CSE

isuruf:
Since Derivatives and Subs are not CSEed due to some issues
derivatives of hankel1 functions and hankel1 functions
end up with different arguments and code generation treats
them seperately. See https://github.com/inducer/sumpy/blob/0fc6535b6ef803e2a8ebd418162274634aead6f9/sumpy/cse.py#L84

Fixes https://github.com/inducer/sumpy/issues/61

### DIFF
--- a/sumpy/symbolic.py
+++ b/sumpy/symbolic.py
@@ -267,9 +267,8 @@ class _BesselOrHankel(sympy.Function):
         if argindex in (1, 3):
             # we are not differentiating w.r.t order or nderivs
             return 0
-        args = list(self.args)
-        args[-1] += 1
-        return self.func(*args)
+        order, z, nderivs = self.args
+        return self.func(order, z, nderivs+1)
 
 
 class BesselJ(_BesselOrHankel):
--- a/sumpy/symbolic.py
+++ b/sumpy/symbolic.py
@@ -241,4 +241,53 @@ class PymbolicToSympyMapperWithSymbols(PymbolicToSympyMapper):
         else:
             self.raise_conversion_error(expr)
 
+    def map_call(self, expr):
+        if expr.function.name == "hankel_1":
+            args = [self.rec(param) for param in expr.parameters]
+            args.append(0)
+            return Hankel1(*args)
+        elif expr.function.name == "bessel_j":
+            args = [self.rec(param) for param in expr.parameters]
+            args.append(0)
+            return BesselJ(*args)
+        else:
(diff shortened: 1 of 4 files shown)


sumpy/symbolic.py
```diff
@@ -241,4 +241,53 @@ def map_subscript(self, expr):
...
+    arguments are (order, z, nderivs)"""
+    nargs = (3,)
+
+    def fdiff(self, argindex=1):
+        if argindex in (1, 3):
+            # we are not differentiating w.r.t order or nderivs
+            return 0
+        args = list(self.args)
+        args[-1] += 1
+        return self.func(*args)
```
inducer, line 272:
```suggestion
        order, z, nderivs = list(self.args)
        return self.func(order, z, nderivs+1)
```

sumpy/symbolic.py
```diff
@@ -241,4 +241,53 @@ def map_subscript(self, expr):
...
+class _BesselOrHankel(sympy.Function):
+    """A symbolic function for BesselJ or Hankel1 functions
+    that keeps track of the derivatives taken of the function
+    arguments are (order, z, nderivs)"""
+    nargs = (3,)
+
+    def fdiff(self, argindex=1):
+        if argindex in (1, 3):
+            # we are not differentiating w.r.t order or nderivs
+            return 0
```
inducer, line 269:
```suggestion
            raise ValueError()
```
?
(Is this code path ever hit? I feel much better about errors than silent garbage data.)

inducer (commented):
Thanks for investigating! Two questions/suggestions, otherwise LGTM!

sumpy/symbolic.py
```diff
@@ -241,4 +241,53 @@ def map_subscript(self, expr):
...
+
+
+import sympy
+
+
+class _BesselOrHankel(sympy.Function):
+    """A symbolic function for BesselJ or Hankel1 functions
+    that keeps track of the derivatives taken of the function
+    arguments are (order, z, nderivs)"""
+    nargs = (3,)
```
inducer, line 264:
```suggestion
    that keeps track of the derivatives taken of the function.
    Arguments are ``(order, z, nderivs)``.
    """
    nargs = (3,)
```

sumpy/symbolic.py
isuruf: Not sure. Let's try

inducer:
Thanks!

PR #65, opened 2021-06-11
Remove BlockIndexRanges and friends from sumpy

alexfikl:
Moving these to pytential, since sumpy just evaluates a linear set of `(tgtindices, srcindices)` doesn't need to know if they're further broken up into blocks or whatnot.

Required by inducer/pytential#92.


alexfikl:
> Adjust branches for pytential downstream CI

I was wondering how to do that. Thanks!

inducer (approved):

inducer:
Thanks!

PR #68, opened 2021-06-24
Introduce an atol parameter to nullspace

isuruf:


sumpy/tools.py
```diff
@@ -737,11 +737,12 @@ def is_obj_array_like(ary):
             or (isinstance(ary, np.ndarray) and ary.dtype.char == "O"))
 
 
-def reduced_row_echelon_form(m):
+def reduced_row_echelon_form(m, atol=0):
     """Calculates a reduced row echelon form of a
     matrix `m`.
 
     :arg m: a 2D :class:`numpy.ndarray` or a list of lists or a sympy Matrix
+    :arg tol: absolute tolerance for values to be considered zero
```
inducer, line 745:
```suggestion
    :arg atol: absolute tolerance for values to be considered zero
```

sumpy/tools.py
```diff
@@ -756,9 +757,11 @@ def reduced_row_echelon_form(m):
             break
         pivot = nrows
         for k in range(index, nrows):
-            if mat[k, i] != 0 and pivot == nrows:
+            if isinstance(mat[k, i], sym.Basic) and not mat[k, i].is_number:
+                continue
```
inducer, line 761: Wouldn't this consider arbitrary symbolic things zero? (Maybe add a comment to explain?

sumpy/tools.py
```diff
@@ -756,9 +757,11 @@ def reduced_row_echelon_form(m):
...
+            if abs(mat[k, i]) > atol and pivot == nrows:
                 pivot = k
-            if abs(mat[k, i]) == 1:
+            if abs(mat[k, i] - 1) <= atol:
                 pivot = k
                 break
         if pivot == nrows:
```
inducer, line 767:
```suggestion
        if pivot == nrows:
            # no nonzero pivot found, next column
```

inducer (commented):
Thanks! A few things below.

sumpy/tools.py
isuruf: Good catch. Will fix it.

inducer:
LGTM, thanks!

PR #76, opened 2021-08-12
Add more docs to P2E classes

alexfikl:
* Adds some docs to the `__call__` methods to describe the inputs.
* Adds support for object arrays.

(Matrix classes were moved to `pytential` to use `P2QBXL` instead)


alexfikl:
@inducer This just adds some docs now, so it should be much less controversial.

inducer:
Thanks!

PR #77, opened 2021-08-13
Fix rscale dtype inference when using object arrays

alexfikl:
Managed to spot an error in #76.

On the other hand, no one is using the object array support in the P2E classes, so I'm fine with deleting it altogether too. It was initially meant for pytential, since all the `nodes()` are object arrays.


inducer:
Thanks for following up. I think it's OK to leave it in in fixed form.

PR #78, opened 2021-08-20
Fix get_args for Stresslet kernel

alexfikl:
Mostly copied form the [ElasticityKernel](https://github.com/inducer/sumpy/blob/a9a6806329cd2043a8d7aff9cd2f7a7131f153eb/sumpy/kernel.py#L747-L752), which already uses `get_all_variables`.

For some context, I was just setting `mu_sym = 1` in the kernel (for non-dimensional equations) and loopy wasn't happy that it got a number as a `ValueArg`.


inducer:
Thanks!

PR #80, opened 2021-09-07
Check AxisTargetDerivative on TargetPointMultiplier

alexfikl:
This tests an `AxisTargetDerivative` on a `TargetPointMultiplier` and it's not quite working as expected. 

At least in the case `AxisTargetDerivative(1, TargetPointMultiplier(0, knl))`, I would expect the target derivative to not do much, since the multiplier is on a different component, so it should be equivalent to `TargetPointMultiplier(0, AxisTargetDerivative(1, knl))`. Am I misunderstanding that?

This is more of a question-PR, so
* Does `TargetPointMultiplier` actually support derivatives?
* If yes, is this the right way to apply / check them?

### DIFF
--- a/sumpy/expansion/multipole.py
+++ b/sumpy/expansion/multipole.py
@@ -91,6 +91,11 @@ class VolumeTaylorMultipoleExpansionBase(MultipoleExpansionBase):
             rscale = 1
 
         base_taker = kernel.get_derivative_taker(bvec, rscale, sac)
+        # Following is a no-op, but reduces the cases we have to handle
+        # in AxisTargetDerivative.postprocess_at_target and
+        # DirectionalTargetDerivative.postprocess_at_target because
+        # the type will always be DifferentiatedExprDerivativeTaker
+        # that we are passing to postprocess_at_target
         base_taker = DifferentiatedExprDerivativeTaker(base_taker,
                 {tuple([0]*self.dim): 1})
         taker = kernel.postprocess_at_target(base_taker, bvec)
--- a/sumpy/expansion/multipole.py
+++ b/sumpy/expansion/multipole.py
@@ -93,7 +93,7 @@ class VolumeTaylorMultipoleExpansionBase(MultipoleExpansionBase):
         base_taker = kernel.get_derivative_taker(bvec, rscale, sac)
         # Following is a no-op, but AxisTargetDerivative.postprocess_at_target and
         # DirectionalTargetDerivative.postprocess_at_target only handles
-        # DifferentiatedExprDerivativeTaker and sympy expression, so we need to
+        # DifferentiatedExprDerivativeTaker and sympy expressions, so we need to
         # make the taker a DifferentitatedExprDerivativeTaker instance.
         base_taker = DifferentiatedExprDerivativeTaker(base_taker,
                 {tuple([0]*self.dim): 1})
--- a/sumpy/expansion/multipole.py
+++ b/sumpy/expansion/multipole.py
@@ -91,11 +91,10 @@ class VolumeTaylorMultipoleExpansionBase(MultipoleExpansionBase):
             rscale = 1
 
         base_taker = kernel.get_derivative_taker(bvec, rscale, sac)
-        # Following is a no-op, but reduces the cases we have to handle
-        # in AxisTargetDerivative.postprocess_at_target and
-        # DirectionalTargetDerivative.postprocess_at_target because
-        # the type will always be DifferentiatedExprDerivativeTaker
-        # that we are passing to postprocess_at_target
+        # Following is a no-op, but AxisTargetDerivative.postprocess_at_target and
+        # DirectionalTargetDerivative.postprocess_at_target only handles
+        # DifferentiatedExprDerivativeTaker and sympy expression, so we need to
+        # make the taker a DifferentitatedExprDerivativeTaker instance.
         base_taker = DifferentiatedExprDerivativeTaker(base_taker,
                 {tuple([0]*self.dim): 1})
         taker = kernel.postprocess_at_target(base_taker, bvec)
--- a/sumpy/expansion/multipole.py
+++ b/sumpy/expansion/multipole.py
@@ -86,10 +86,13 @@ class VolumeTaylorMultipoleExpansionBase(MultipoleExpansionBase):
                 rscale, (1,), sac=sac)
 
     def evaluate(self, kernel, coeffs, bvec, rscale, sac=None):
+        from sumpy.tools import DifferentiatedExprDerivativeTaker
         if not self.use_rscale:
             rscale = 1
 
         base_taker = kernel.get_derivative_taker(bvec, rscale, sac)
+        base_taker = DifferentiatedExprDerivativeTaker(base_taker,
+                {tuple([0]*self.dim): 1})
         taker = kernel.postprocess_at_target(base_taker, bvec)
 
         result = []
--- a/sumpy/kernel.py
+++ b/sumpy/kernel.py
@@ -239,32 +239,10 @@ class Kernel:
         The typical use of this function is to apply target-variable
         derivatives to the kernel.
         """
-        from sumpy.tools import (ExprDerivativeTaker,
-            DifferentiatedExprDerivativeTaker)
-        expr_dict = {(0,)*self.dim: 1}
-        expr_dict = self.get_derivative_transformation_at_target(expr_dict)
-        if isinstance(expr, ExprDerivativeTaker):
-            return DifferentiatedExprDerivativeTaker(expr, expr_dict)
-
-        result = 0
-        for mi, coeff in expr_dict.items():
-            result += coeff * self._diff(expr, bvec, mi)
-        return result
+        return expr
 
     def get_derivative_transformation_at_source(self, expr_dict):
         r"""Get the derivative transformation of the expression at source
-        represented by the dictionary expr_dict which is mapping from multi-index
-        `mi` to coefficient `coeff`.
-        Expression represented by the dictionary `expr_dict` is
-        :math:`\sum_{mi} \frac{\partial^mi}{x^mi}G * coeff`. Returns an
-        expression of the same type.
-
-        This function is meant to be overridden by child classes where necessary.
-        """
-        return expr_dict
-
-    def get_derivative_transformation_at_target(self, expr_dict):
-        r"""Get the derivative transformation of the expression at target
         represented by the dictionary expr_dict which is mapping from multi-index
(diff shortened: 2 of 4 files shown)


isuruf:
I didn't intend to have `AxisTargetDerivative(1, TargetPointMultiplier(0, knl))` to work. We should probably raise an error for that.

alexfikl:
> I didn't intend to have `AxisTargetDerivative(1, TargetPointMultiplier(0, knl))` to work. We should probably raise an error for that.

I came across this when trying to compute the velocity gradient using the `StokesletWrapperTornberg` in [here](https://github.com/isuruf/pytential/blob/c794ad04decd1256f08aabf108e5e293d4c323e2/pytential/symbolic/stokes.py#L517-L520). It would be nice to get that working, but if not, I agree that raising an error somewhere would be helpful!

isuruf:
Ah, I'll prepare a PR for this. Thanks for the test case.

alexfikl:
> Ah, I'll prepare a PR for this. Thanks for the test case.

That would be awesome! Thank you.

alexfikl:
That seems to have fixed everything very nicely! Can confirm that computing velocity gradients on the pytential side also works with this little patch

```diff
diff --git a/pytential/symbolic/stokes.py b/pytential/symbolic/stokes.py
index aeff997..e539971 100644
--- a/pytential/symbolic/stokes.py
+++ b/pytential/symbolic/stokes.py
@@ -524,7 +524,12 @@ class StressletWrapperTornberg(StressletWrapperBase):
                     qbx_forced_limit=qbx_forced_limit)
 
                 if i == j:
-                    sym_expr[i] += sym.IntG(target_kernel=self.kernel,
+                    target_kernel = self.kernel
+                    for deriv_dir in extra_deriv_dirs:
+                        target_kernel = AxisTargetDerivative(
+                                deriv_dir, target_kernel)
+
+                    sym_expr[i] += sym.IntG(target_kernel=target_kernel,
                         source_kernels=common_source_kernels,
                         densities=densities,
                         qbx_forced_limit=qbx_forced_limit
```

sumpy/expansion/multipole.py
```diff
@@ -86,10 +86,13 @@ def coefficients_from_source(self, kernel, avec, bvec, rscale, sac=None):
                 rscale, (1,), sac=sac)
 
     def evaluate(self, kernel, coeffs, bvec, rscale, sac=None):
+        from sumpy.tools import DifferentiatedExprDerivativeTaker
         if not self.use_rscale:
             rscale = 1
 
         base_taker = kernel.get_derivative_taker(bvec, rscale, sac)
+        base_taker = DifferentiatedExprDerivativeTaker(base_taker,
+                {tuple([0]*self.dim): 1})
```
inducer, line 95: Why is this helpful? (IIUC this should be equivalent to calling the underlying derivative taker?) If not, add a comment to explain.

sumpy/kernel.py
```diff
@@ -239,17 +239,7 @@ def postprocess_at_target(self, expr, bvec):
...
-        expr_dict = {(0,)*self.dim: 1}
-        expr_dict = self.get_derivative_transformation_at_target(expr_dict)
-        if isinstance(expr, ExprDerivativeTaker):
-            return DifferentiatedExprDerivativeTaker(expr, expr_dict)
-
-        result = 0
-        for mi, coeff in expr_dict.items():
-            result += coeff * self._diff(expr, bvec, mi)
-        return result
+        return expr
```
inducer, line 242: This is strange to me... wouldn't this default implementation risk invalidly skipping derivatives?

sumpy/kernel.py
```diff
@@ -263,18 +253,6 @@ def get_derivative_transformation_at_source(self, expr_dict):
         """
         return expr_dict
 
-    def get_derivative_transformation_at_target(self, expr_dict):
```
inducer, line 266: What's happening here? Do you intend for this to remain part of the `Kernel` interface? If so, I think it'd be good to leave an `@abstractmethod` behind. In that case, it should probably also be roped into the class documentation.

sumpy/kernel.py
```diff
@@ -961,8 +939,8 @@ def get_expression(self, scaled_dist_vec):
     def get_derivative_transformation_at_source(self, expr_dict):
         return self.inner_kernel.get_derivative_transformation_at_source(expr_dict)
 
-    def get_derivative_transformation_at_target(self, expr_dict):
-        return self.inner_kernel.get_derivative_transformation_at_target(expr_dict)
+    def postprocess_at_target(self, expr, bvec):
+        return self.inner_kernel.postprocess_at_target(expr, bvec)
```
inducer, line 943: Again, super confused by this change...

inducer (commented):
I think I need help navigating this. I'm not sure I'm able to make sense of this.

sumpy/expansion/multipole.py
isuruf: It's for easier code to maintain. If you look at https://github.com/inducer/sumpy/blob/97fdea77d71a7e32e0c8eaa7b7beb885ea9e1ef4/sumpy/kernel.py#L1035 there are two code branches.
1. DifferentiatedExprDerivativeTaker instances
2. Regular sympy expressions

If I didn't have this here, we would have a third code branch,

3. DerivativeTaker instances that are not DifferentiatedExprDerivativeTaker instances

sumpy/expansion/multipole.py
```diff
@@ -86,10 +86,13 @@ def coefficients_from_source(self, kernel, avec, bvec, rscale, sac=None):
                 rscale, (1,), sac=sac)
 
     def evaluate(self, kernel, coeffs, bvec, rscale, sac=None):
+        from sumpy.tools import DifferentiatedExprDerivativeTaker
         if not self.use_rscale:
             rscale = 1
 
         base_taker = kernel.get_derivative_taker(bvec, rscale, sac)
+        base_taker = DifferentiatedExprDerivativeTaker(base_taker,
```
inducer, line 99:
```suggestion
        # Make it so that postprocess_at_target can assume that the
        # "expression" passed to it is either just a sympy expression
        # or a DifferentiatedExprDerivativeTaker.
        base_taker = DifferentiatedExprDerivativeTaker(base_taker,
```

sumpy/expansion/multipole.py
inducer: Thx!

sumpy/kernel.py
```diff
@@ -239,17 +239,7 @@ def postprocess_at_target(self, expr, bvec):
         The typical use of this function is to apply target-variable
         derivatives to the kernel.
         """
```
inducer, line 241:
```suggestion
        
        :arg expr: may be a :class:`sympy.Expr` or a
            :class:`sumpy.tools.DifferentiatedExprDerivativeTaker`.
        """
```

sumpy/expansion/multipole.py
inducer: (Thanks for also taking a stab at this explanation. Maybe make one that takes the best aspects from both of ours?)

sumpy/tools.py
```diff
@@ -430,6 +431,26 @@ def diff(self, mi, save_intermediate=lambda x: x):
 
         return result * save_intermediate(1 / self.taker.rscale ** max_order)
 
+
+def diff_transformation(derivative_transformation, variable_idx, variables):
```
inducer, line 435:
```suggestion
def diff_derivative_coeff_dict(derivative_coeff_dict, variable_idx, variables):
```
(I don't know really... these dictionaries are starting to proliferate, indicating that they're quite useful. I'm not sure I'm loving "derivative transformation as a name for them. Maybe we should have a separate type for them, even if it's just a one-attribute wrapper around the dictionary?)

sumpy/tools.py
```diff
@@ -430,6 +431,26 @@ def diff(self, mi, save_intermediate=lambda x: x):
...
+    """Differentiate a derivative transformation dictionary using the
+    variable given by **variable_idx** and return a new derivative transformation
+    dictionary
+    """
+    new_transformation = defaultdict(lambda: 0)
```
inducer, line 440:
```suggestion
    new_derivative_coeff_dict = defaultdict(lambda: 0)
```

sumpy/tools.py
```diff
@@ -430,6 +431,26 @@ def diff(self, mi, save_intermediate=lambda x: x):
...
+    for mi, coeff in derivative_transformation.items():
+        # In the case where we have x * u.diff(x), the result should
+        # be x.diff(x) + x * u.diff(x, x)
+        # Calculate the first term by differentiating the coefficients
+        new_coeff = sym.sympify(coeff).diff(variables[variable_idx])
+        if new_coeff != 0:
+            new_transformation[mi] += new_coeff
+        # Next calculate the second term by differentitating the derivatives
```
inducer, line 448:
```suggestion
        # Next calculate the second term by differentiating the derivatives
```

sumpy/tools.py
```diff
@@ -430,6 +431,26 @@ def diff(self, mi, save_intermediate=lambda x: x):
...
+    dictionary
+    """
+    new_transformation = defaultdict(lambda: 0)
+    for mi, coeff in derivative_transformation.items():
+        # In the case where we have x * u.diff(x), the result should
+        # be x.diff(x) + x * u.diff(x, x)
+        # Calculate the first term by differentiating the coefficients
+        new_coeff = sym.sympify(coeff).diff(variables[variable_idx])
+        if new_coeff != 0:
+            new_transformation[mi] += new_coeff
```
inducer, line 447:
```suggestion
        new_transformation[mi] += new_coeff
```

inducer (commented):
Thanks! This looks good. Just a few style/doc complaints, as usual. :)

sumpy/expansion/multipole.py
```diff
@@ -86,10 +86,17 @@ def coefficients_from_source(self, kernel, avec, bvec, rscale, sac=None):
...
 
     def evaluate(self, kernel, coeffs, bvec, rscale, sac=None):
+        from sumpy.tools import DifferentiatedExprDerivativeTaker
         if not self.use_rscale:
             rscale = 1
 
         base_taker = kernel.get_derivative_taker(bvec, rscale, sac)
+        # Following is a no-op, but AxisTargetDerivative.postprocess_at_target and
+        # DirectionalTargetDerivative.postprocess_at_target only handles
+        # DifferentiatedExprDerivativeTaker and sympy expression, so we need to
```
isuruf, line 96:
```suggestion
        # DifferentiatedExprDerivativeTaker and sympy expressions, so we need to
```

inducer:
Looks great. Thanks!

PR #81, opened 2021-09-13
Remove deprecation of specifying the direction vector name

isuruf:


isuruf:
One use-case would be to specify two different source direction vectors.

inducer:
Thx!

PR #86, opened 2021-10-09
Add SpatialConstant

isuruf:
Partially fixes https://github.com/inducer/pytential/issues/116

### DIFF
--- a/sumpy/symbolic.py
+++ b/sumpy/symbolic.py
@@ -20,6 +20,14 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
+__doc__ = """
+
+ Symbolic Tools
+ ==============
+
+ .. autoclass:: SpatialConstant
+"""
+
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


sumpy/symbolic.py
```diff
@@ -243,6 +244,33 @@ def find_power_of(base, prod):
     return result[power]
 
 
+class SpatialConstant(prim.Variable):
```
inducer, line 247: I'd prefer if this had its own `mapper_method`, instead of piggybacking off of `map_variable` and using `isinstance`.

isuruf: What do you think about moving this to `pymbolic`? If we don't move this `pymbolic`, issue is that, I have to override a lot of mappers of pymbolic in both sumpy and pytential with `map_spatial_constant=map_variable`.

sumpy/symbolic.py
```diff
@@ -243,6 +244,30 @@ def find_power_of(base, prod):
...
```
inducer, line 247: Hook this into the documentation and write a docstring explaining the motivation.

isuruf: Done in 8e7435a. Ready for review

PR #100, opened 2022-03-06
Fix diff with sympy 1.10

alexfikl:
Just updated and this seems to fail
```
>> sp.sympify("tau", strict=True)
sympy.core.sympify.SympifyError: SympifyError: 'tau'
```
Here it gets called by `diff` somewhere (`subs` still works just fine, but changed it just to be safe).


isuruf (approved):

inducer:
Thx!

alexfikl:
For reference, the extra `sympify` seems to have been introduced in sympy/sympy#23162.

PR #105, opened 2022-04-06
Optimize P2P for GPU

isuruf:
This calculates the maximum number of points in one box and each target is assigned a work item.
If the particular target box doesn't the maximum, that work item does nothing (except to bring sources to local memory)
Before calculating each work-item brings one source point and charge to local memory.
Then each work item calculates the value for that target.

cf. https://gitlab.tiker.net/inducer/sumpy/-/merge_requests/152


sumpy/p2p.py
```diff
@@ -480,30 +487,51 @@ def get_kernel(self):
...
-                    for itgt
-                    for isrc
-                        <> d[idim] = \
-                            targets[idim, itgt] - sources[idim, isrc]
+                    <> cond_isrc = ipoint < isrc_end - isrc_start
+                    if cond_isrc
+                      local_isrc[idim, ipoint] = sources[idim,
+                          ipoint + isrc_start]  {id=load_src, dup=idim}
+                      local_isrc_strength[istrength, ipoint] = strength[istrength,
+                          ipoint + isrc_start]  {id=load_charge}
```
inducer, line 506: Can't `add_prefetch` add this?

sumpy/fmm.py
```diff
@@ -674,7 +682,8 @@ def eval_direct(self, target_boxes, source_box_starts,
                 source_box_lists=source_box_lists,
                 strength=src_weight_vecs,
                 result=pot,
-
+                max_npoints_in_one_box=max(self.get_max_nsources_in_one_box(queue),
+                    self.get_max_ntargets_in_one_box(queue)),
```
inducer, line 686: Not sure I'm a fan of this being recomputed on an ongoing basis...

isuruf: Fixed in 447e5eb

sumpy/p2p.py
isuruf: I tried doing that, but I'm not sure I was doing it correct.
I get a,
```
RuntimeError: Inames 'idim, isrc' require more than one leaf domain, which makes the domain change that is part of your current operation ambiguous.
```

with `knl = lp.add_prefetch(knl, "sources")`.
Any ideas?

isuruf:
Looks like pocl miscompiles the prefetch for CPU device. pocl-cuda, nvidia opencl, and oclgrind works just fine.
Maybe I should disable the prefetching for CPU devices.

sumpy/p2p.py
inducer: If you duplicate the `idim` to the point where it's only used by this statement, this should resolve itself.

isuruf:
```
>               raise RuntimeError("no invocations of '%s' found" % subst_name)
E               RuntimeError: no invocations of 'sources_fetch_rule' found

../loopy/loopy/transform/precompute.py:550: RuntimeError
```

isuruf:
I wrote a separate loopy kernel for the GPU, because I couldn't figure out how to go from the naive CPU version to the GPU using existing loopy transforms.

sumpy/p2p.py
```diff
@@ -459,19 +459,42 @@ def get_kernel(self):
...
-            + ["""
+            "{[idim]: 0 <= idim < dim}",
+            "{[istrength]: 0 <= istrength < nstrengths}",
+            "{[isrc]: isrc_start <= isrc < isrc_end}"
+        ]
+
+        src_outer_limit = (max_nsources_in_one_box - 1) // nsplit
+        tgt_outer_limit = (max_ntargets_in_one_box - 1) // nsplit
+
+        if gpu:
```
inducer, line 478: If you run the GPU kernel on the CPU, does it do worse? (What I'm aiming for is to only have one of them.)

inducer:
Out of curiosity: How well does the GPU kernel do? (Given some model, say.)

sumpy/p2p.py
isuruf: Pocl miscompiles the GPU kernel and there are NaNs as output in some tests.

inducer: * Only pocl-cpu or all pocls?
* Could you add a comment to explain why we have both?

isuruf: pocl-cpu only. Added a comment in 955d279f68daa4d5103b1165ef8c70f075b9dd6f

isuruf:
> Out of curiosity: How well does the GPU kernel do? (Given some model, say.)

On a Titan V it does 2.0 TFLOPs for Laplace 3D with 40 max points per box whereas the peak is 6.8 TFLOPs.

inducer:
Super cool, LGTM! Thanks for working on this.

PR #106, opened 2022-04-11
m2l: add dependencies for inlined e2e kernel

alexfikl:
This was generating a lot of warnings like
```
loopy/kernel/creation.py:1911: LoopyWarning: in kernel m2l_using_translation_classes_dependent_data: 
The single-writer dependency heuristic added dependencies on instruction ID(s)
'read_src_ibox, translation_offset' to instruction ID 'e2e_insn_3' after
kernel creation is complete. This is deprecated and may stop working in the
future. To fix this, ensure that instruction dependencies are added/resolved 
as soon as possible, ideally at kernel creation time. (add 
'single_writer_after_creation' to silenced_warnings kernel attribute to disable)
```

Not sure if this is the right way to add the required dependencies, so better suggestions welcome!

### DIFF
--- a/sumpy/e2e.py
+++ b/sumpy/e2e.py
@@ -424,8 +424,9 @@ class M2LUsingTranslationClassesDependentData(E2EFromCSR):
                                 {id=read_src_ibox}
                         <> translation_class = \
                                 m2l_translation_classes_lists[isrc_box]
-                        <> translation_class_rel = translation_class - \
-                                                    translation_classes_level_start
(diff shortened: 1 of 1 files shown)


isuruf (approved):

inducer:
> Not sure if this is the right way to add the required dependencies, so better suggestions welcome!

It's based on statement IDs (not tags), which isn't ideal, but it'll do for now. It's certainly better than the previous state!

inducer:
Thanks!

PR #26, opened 2021-02-04
Custom derivative takers

isuruf:
- [x] Wait for https://github.com/inducer/pytential/pull/65

PR #108, opened 2022-04-17
Fix HelmholtzDerivativeTaker

### LINKED ISSUE #107: test_translations fails when run with 3D kernels
Currently only 2D kernels are tested and when 3D kernels are tested the convergence
verifier fails.

isuruf:
Fixes #107 
Regression from #26 

Also adds a test for 3D kernels in test_translations.

### DIFF
--- a/sumpy/tools.py
+++ b/sumpy/tools.py
@@ -384,17 +384,15 @@ class HelmholtzDerivativeTaker(RadialDerivativeTaker):
             # See https://dlmf.nist.gov/10.6.E6
             # and https://dlmf.nist.gov/10.6#E1
             k = self.orig_expr.args[1] / self.r
-            expr = -  2 * (q - 1) * self.diff(mi, q - 1)
-            expr += - k**2 * self.diff(mi, q - 2)
-            expr /= self.r**2
+            expr = (-2*(q - 1) * self.diff(mi, q - 1)
+                    - k**2 * self.diff(mi, q - 2)) / self.r**2
         else:
             # See reference [Tausch2003] in RadialDerivativeTaker.diff
             # Note that there is a typo in the paper where
             # -k**2/r is given instead of -k**2/r**2.
             k = (self.orig_expr * self.r).args[-1] / sym.I / self.r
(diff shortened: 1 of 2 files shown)


inducer:
Thanks!

sumpy/tools.py
```diff
@@ -388,9 +389,11 @@ def diff(self, mi, q=0):
             expr /= self.r**2
         else:
             # See reference [Tausch2003] in RadialDerivativeTaker.diff
+            # Note that there is a typo in the paper where
+            # -k**2/r is given instead of -k**2/r**2.
```
inducer, line 393: Nice find! :tada:

test/test_kernels.py
```diff
@@ -525,7 +525,10 @@ def test_translations(ctx_factory, knl, local_expn_class, mpole_expn_class):
 
     del eval_offset
 
-    orders = [3, 4, 5]
+    if knl.dim == 2:
+        orders = [2, 3, 4]
```
inducer, line 529: How come?

sumpy/tools.py
```diff
@@ -392,8 +392,9 @@ def diff(self, mi, q=0):
             # Note that there is a typo in the paper where
             # -k**2/r is given instead of -k**2/r**2.
             k = (self.orig_expr * self.r).args[-1] / sym.I / self.r
-            expr = -(2*q - 1)/self.r**2 * self.diff(mi, q - 1)
-            expr += -k**2 / self.r**2 * self.diff(mi, q - 2)
+            expr = -(2*q - 1) * self.diff(mi, q - 1)
+            expr += -k**2 * self.diff(mi, q - 2)
+            expr /= self.r**2
```
inducer, line 397: Want to write this as a single expression? Less error-prone IMO.

test/test_kernels.py
isuruf: 4->5 is problematic for most of 2D kernels like 2->3 is problematic for 3D kernels.

inducer: Fair enough. :shrug:

inducer:
Thanks!

PR #109, opened 2022-04-17
Add test for compressed Helmholtz error

isuruf:
Checks that the error is of the same order as the truncation error.

### DIFF
--- a/test/test_kernels.py
+++ b/test/test_kernels.py
@@ -866,6 +866,81 @@ def test_m2l_toeplitz():
 # }}}
 
 
+# {{{ test_m2m_compressed
+
+@pytest.mark.parametrize("dim", [2, 3])
+@pytest.mark.parametrize("order", [2, 4, 6])
+def test_m2m_compressed_error_helmholtz(ctx_factory, dim, order):
+    import sumpy.toys as t
+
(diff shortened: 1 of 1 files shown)


test/test_kernels.py
```diff
@@ -866,6 +866,81 @@ def test_m2l_toeplitz():
...
+    from pytools.convergence import EOCRecorder
+    eoc_rec = EOCRecorder()
+    h_values = 2.0**np.arange(-7, -3)
+    for h in h_values:
+        sources = (h*(-0.5+sources_grid.astype(np.float64)) + mpole_center)
+        second_center = mpole_center + h/2
+        furthest_source = np.max(np.abs(sources - mpole_center))
+        m2m_vals = [0, 0]
+        for i, (mpole_expn_class, local_expn_class) in \
+                enumerate(zip(mpole_expn_classes, local_expn_classes)):
```
inducer, line 910: Let pytest parametrize this?

isuruf: We are calculating the error between compressed vs full with full regarded as the reference value. So we need both to calculate the error.

inducer: Whoops. Good point.

inducer:
Thanks!

PR #110, opened 2022-04-19
Remove uses of pytools.factorial

alexfikl:
For context, just fixing these because `pytential` started showing loads of deprecation warnings :cry:

### DIFF
--- a/contrib/translations/PDE-reduction-symbolic.ipynb
+++ b/contrib/translations/PDE-reduction-symbolic.ipynb
@@ -15,9 +15,7 @@
     "from pytools import add_tuples\n",
     "\n",
     "from sumpy.expansion.local import VolumeTaylorLocalExpansion\n",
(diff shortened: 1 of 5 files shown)


contrib/translations/PDE-reduction-symbolic.ipynb
```diff
@@ -48,20 +46,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from pytools import factorial\n",
-    "\n",
-    "def mi_factorial(n):\n",
-    "    return np.prod([factorial(n1) for n1 in n])"
```
alexfikl, line 54: The notebook had this in, but it doesn't seem used.

I also updated it so it actually runs, `mpole_expn_reduced.derivative_wrangler` didn't seem to exist anymore. Not sure it's giving the correct results though.

inducer:
Thx!

PR #111, opened 2022-04-19
Fix, test order varying with level

inducer:
cc @gaohao95


inducer:
@gaohao95 Turns out the error you got was spurious. This PR applies a small fix and removes the check.

PR #116, opened 2022-05-08
Use unicode in DirectionalDerivative.__str__

alexfikl:
Thoughts? Was looking at too many of these in inducer/pytential#152 and this looked prettier..


inducer:
:shrug: Let's try it and see if we like it. Thanks!

PR #117, opened 2022-05-10
Compute translation_classes_data if not given

isuruf:
Adds a `_disable_translation_classes` hidden option so that we can check the non translation classes kernel.

### DIFF
--- a/sumpy/fmm.py
+++ b/sumpy/fmm.py
@@ -315,7 +315,8 @@ class SumpyExpansionWrangler(ExpansionWranglerInterface):
             kernel_extra_kwargs=None,
             self_extra_kwargs=None,
             translation_classes_data=None,
-            preprocessed_mpole_dtype=None):
+            preprocessed_mpole_dtype=None,
+            *, _disable_translation_classes=False):
         super().__init__(tree_indep, traversal)
         self.issued_timing_data_warning = False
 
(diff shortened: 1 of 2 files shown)


inducer:
Thx!

sumpy/fmm.py
```diff
@@ -353,27 +354,14 @@ def __init__(self, tree_indep, traversal, dtype, fmm_level_to_order,
...
-            else:
-                self.supports_translation_classes = True
-        else:
+        if _disable_translation_classes or not base_kernel.is_translation_invariant:
             self.supports_translation_classes = False
+        else:
+            if translation_classes_data is None:
+                with cl.CommandQueue(self.tree_indep.cl_context) as queue:
+                    translation_classes_data = SumpyTranslationClassesData(queue,
+                        self.traversal)
```
alexfikl, line 363: Why does this work? Doesn't the queue get destroyed when it exits the context manager leaving `SumpyTranslationClassesData` holding an invalid reference?

isuruf: Yes, the queue completes and gets destroyed, but the data itself doesn't get destroyed. i.e. the data has `queue=None`. When a new queue needs this data, they are accessed with `.with_queue(queue)`

alexfikl: https://github.com/inducer/sumpy/blob/a2c830abed936e42424aad025b6d44bd82bb7750/sumpy/fmm.py#L244-L249
Hm, but the constructor doesn't do any work. The actual work is being done when calling some of those memoized methods below, which should be after the context manager exists.

alexfikl: The tests don't seem to crash and burn, so this is probably just some misunderstanding on my side. Hm..

isuruf: Hmm, you are right. I guess since the constructor keeps a reference to the queue, it doesn't get destroyed.

alexfikl: https://github.com/inducer/pyopencl/blob/175060ab392e40866dc1291eda13fec724ac7acc/pyopencl/__init__.py#L707-L708

Yeah, you're right! `__exit__` doesn't destroy it, it just calls finish, so should be ok!

inducer: Good catch, @alexfikl! I was under the mistaken assumption that the work got done by the constructor.

The context manager does not destroy the queue:

https://github.com/inducer/pyopencl/blob/175060ab392e40866dc1291eda13fec724ac7acc/pyopencl/__init__.py#L704-L708=

But continuing to hold a reference to the queue past `__exit__` is definitely not the intended use of the interface.

(Fun fact: Neither CL nor PyOpenCL allows destroying a queue, only releasing a reference to it.)

inducer: I'll try to come up with a way to make PyOpenCL warn if a reference to the queue continues to be held past `__exit__`.

alexfikl: That would be really cool!

isuruf: I changed the implementation to build the data objects eagerly instead of lazily.

alexfikl: Can't speak for the eager vs lazy, but the queue stuff makes sense to me now :+1:

inducer: https://github.com/inducer/pyopencl/pull/561

inducer:
LGTM, thanks!

PR #119, opened 2022-05-19
M2LTranslationBase: mark immutable, fix equality comparison

inducer:
Without this, sumpy cache retrievals never succeed, because this comparison

https://github.com/inducer/sumpy/blob/daa4118947d5d4a1b892280720d48dde611288e2/sumpy/expansion/local.py#L78=

always returns `False`.


isuruf (approved):
Thanks

PR #122, opened 2022-06-06
4r/order for biharmonic

isuruf:


sumpy/fmm.py
```diff
@@ -323,8 +323,14 @@ def level_to_rscale(self, level):
         # YALE UNIV NEW HAVEN CT DEPT OF COMPUTER SCIENCE, 1988.
         # rscale that we use in sumpy is the inverse of the scaling used in the
         # paper and therefore we should use r / order. However empirically
-        # we have observed that 2r / order is better for numerical stability.
-        return r * 2 / order
+        # we have observed that 2r / order is better for numerical stability
+        # for Laplace and 4r / order for biharmonic kernel.
```
inducer, line 327: Either make a test to reproduce or diagnose this at the coefficient magnitude level, or describe precisely (with commit hashes) what the observation was.

isuruf: Added a test in f7236d6

inducer:
LGTM, thanks!

Issue #127, opened 2022-08-01
VkFFT fails on GPU


inducer:
Gitlab CI run: https://gitlab.tiker.net/inducer/sumpy/-/jobs/439559

@isuruf, could you take a look?

isuruf:
The error is weird. `pyopencl._cl.RuntimeError: clGetEventProfilingInfo failed: PROFILING_INFO_NOT_AVAILABLE`
This error should be reproducible and yet I can't reproduce locally.

inducer:
Did you run on the pocl+cuda device?

isuruf:
Yes, tried `pocl:cuda`, `pocl:pthread` and `nvidia` all of which passes the test.

isuruf:
Nvm, I can reproduce

PR #128, opened 2022-08-02
Fix order in events for AggregateProfilingEvent

isuruf:
AggregateProfilingEvent assumes that the last event depends on all the others.

Fixes https://github.com/inducer/sumpy/issues/127


inducer:
Thanks!

Issue #129, opened 2022-08-05
Funny stuff happening since FFT/vkfft merge


inducer:
Issues showed up after #114:

- Possibly slower
- Missing global barrier in some cases
- Crashes, e.g. https://github.com/inducer/pytential/runs/7625741382?check_suite_focus=true

Fairly unspecific at this point.

Pointed out by @alexfikl.

cc @isuruf

inducer:
Pytential's Mac CI on `main` is failing now, too: https://github.com/inducer/pytential/runs/7694158164?check_suite_focus=true#step:3:1057

A number of @alexfikl's PRs had picked up similar failures, and this establishes that the issue is not specific to them.

I conjecture that this is connected to the #114 merge. To help back up that conjecture, here's a run without that merge: https://github.com/inducer/pytential/pull/175. If there are no failures there, then that is further evidence that #114 is causing problems.

The CI runs don't necessarily back up the slowness claim.
- Here's a CI summary page pre-merge: https://github.com/inducer/pytential/actions/runs/2724593425
- and post merge: https://github.com/inducer/pytential/actions/runs/2770799383

At least the overall CI times are relatively similar. @alexfikl, could you provide a way to reproduce your Stokes run that got substantially slower?

The missing barrier is something I encountered in work with @rckirby. I still have to work out what's happening, I will report back here.

inducer:
The first run of https://github.com/inducer/pytential/pull/175 passed. I've just started another run as an additional data point.

inducer:
Second run passed as well.

alexfikl:
Just to add more information here from my side.

First, for the slowdown, I noticed that on the 3D Stokes operator from inducer/pytential#29 (slightly modified and merged with `main`). That uses the Laplace formulation from
https://github.com/isuruf/pytential/blob/feec35e928767dd6add0cba55883cd15bed93445/pytential/symbolic/stokes.py#L578-L639
I'll try to reproduce something and come back with some numbers.

Then, for the warnings on `sumpy`, they were
```
[...]

PR #131, opened 2022-08-15
loop-y kernel for m2l preprocess

isuruf:


isuruf:
@alexfikl, can you check to see if the slowness is fixed now?

alexfikl:
I just ran the script from https://github.com/inducer/sumpy/issues/129#issuecomment-1207257972 and it's still just as slow.
```
nice python -O examples/benchmark-stokes-single-layer.py
```

This is on `dunkel` with all the repos (`loopy`, etc.) on latest `main`.

inducer:
Thx!

PR #133, opened 2022-08-17
Disable global barrier check in m2l preprocess

inducer:
Here's another one similar to #132.

cc @isuruf @alexfikl


inducer:
cc @rkirby

isuruf:
https://github.com/inducer/sumpy/pull/131 is an alternative to this.

alexfikl:
Can confirm that #131 also does a nice job at fixing this issue!
```
133:	elapsed: 427.10s wall 0.98x CPU
131: 	elapsed: 372.73s wall 1.04x CPU
```
(this is just one run on `dunkel` using https://github.com/inducer/sumpy/issues/129#issuecomment-1207257972, so not rigorous, but still)

inducer:
Sounds good. Let's go with #131 instead.

PR #135, opened 2022-08-20
Use arraycontext in tests

alexfikl:
Ports all the tests to the `arraycontext` infrastructure.

This is not on top of #118 because it can go in without the rest of `sumpy` being ported.

### DIFF
--- a/test/test_fmm.py
+++ b/test/test_fmm.py
@@ -20,42 +20,47 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-
+import pytest
 import sys
 import os
-from unittest.mock import patch
+from functools import partial
+
 import numpy as np
 import numpy.linalg as la
-import pyopencl as cl
-from pyopencl.tools import (  # noqa
-        pytest_generate_tests_for_pyopencl as pytest_generate_tests)
-from sumpy.kernel import (LaplaceKernel, HelmholtzKernel, YukawaKernel,
+
+from arraycontext import pytest_generate_tests_for_array_contexts
+from sumpy.array_context import (                                 # noqa: F401
+        PytestPyOpenCLArrayContextFactory, _acf)
+
+from sumpy.kernel import (
+    LaplaceKernel,
+    HelmholtzKernel,
+    YukawaKernel,
     BiharmonicKernel)
 from sumpy.expansion.multipole import (
     VolumeTaylorMultipoleExpansion,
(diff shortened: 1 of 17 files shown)


alexfikl:
After writing most of this, I realized that there's some small conflicts with #118. I'm happy to resolve those if this gets in.

test/test_fmm.py
```diff
@@ -691,27 +682,24 @@ def test_sumpy_target_point_multiplier(ctx_factory, deriv_axes):
...
     rel_err = la.norm(pot0 - ref_pot) / la.norm(ref_pot)
     logger.info("order %d -> relative l2 error: %g", order, rel_err)
 
     assert np.isclose(rel_err, 0, atol=1e-5)
 
+# }}}
 
-# You can test individual routines by typing
-# $ python test_fmm.py 'test_sumpy_fmm(cl.create_some_context, LaplaceKernel(2),
-#       VolumeTaylorLocalExpansion, VolumeTaylorMultipoleExpansion, False, False)'
```
isuruf, line 708: Can you keep these lines and change `cl.create_some_context` to `_acf`? I use this quite frequently.

alexfikl: Yeah, definitely!

alexfikl:
@inducer This should be ready for a look! 

Most of it is just tedious, but there are a few things worth mentioning:
* This inherits the array context from `boxtree.PyOpenCLArrayContext`.
* There were a couple of tests that used `np.random` and failed when changing the seed.

test/test_fmm.py
alexfikl: This should be back in the latest version, plus updated the comment in the other test files as well.

Thanks for taking a look and let me know if you spot anything else that's fishy!

inducer:
> * There were a couple of tests that used `np.random` and failed when changing the seed in `test_kernels.py`.

Do you remember which ones those were? Were they just random tolerance botches or something you thought was indicative of a deeper issue?

inducer:
This looks good to me, and it's an all-around improvement. Thanks!

alexfikl:
> Do you remember which ones those were? Were they just random tolerance botches or something you thought was indicative of a deeper issue? (If the latter, could you file an issue?)

Looking through the failed CIs (e.g. https://github.com/inducer/sumpy/runs/7933601203?check_suite_focus=true)

* `test_kernels::test_translations`: failed with `assert ([-0.19, -0.8] < -0.2).all()`
* `test_kernels::test_p2e2p`: failed with `assert 3.3 > 3.5`

so I would say it's just a case of fixing the tolerances to the previous seeds, but I don't know enough about what the tests are actually testing to say.

PR #137, opened 2022-08-22
Add distributed wrangler

gaohao95:
Needs:
- [x] Depends on https://github.com/inducer/boxtree/pull/55.
- [x] Point `requirements.txt` back to boxtree main

### DIFF
--- a/.test-conda-env-py3.yml
+++ b/.test-conda-env-py3.yml
@@ -16,3 +16,4 @@ dependencies:
 - pyfmmlib
 - pyrsistent
 - pyvkfft
+- mpi4py
--- a/sumpy/distributed.py
+++ b/sumpy/distributed.py
@@ -1,0 +1,98 @@
+__copyright__ = "Copyright (C) 2022 Hao Gao"
+
+__license__ = """
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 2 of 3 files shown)


sumpy/fmm.py
```diff
@@ -1083,6 +1083,95 @@ def finalize_potentials(self, potentials, template_ary):
 # }}}
 
 
+# {{{ distributed expansion wrangler
+
+try:
+    from boxtree.distributed.calculation import DistributedExpansionWrangler
+    is_distributed_wrangler_available = True
+except ModuleNotFoundError:
+    is_distributed_wrangler_available = False
```
inducer, line 1092: Move to separate file instead.

inducer:
LGTM, thanks!

.test-conda-env-py3.yml
```diff
@@ -16,3 +16,4 @@ dependencies:
 - pyfmmlib
 - pyrsistent
 - pyvkfft
+- mpi4py
```
isuruf, line 19: To be explicit about the test dependency
```suggestion
- mpi4py
- platformdirs
```

inducer: Good catch. Sorry this missed the merge! Is there something like `test_requires` in `setup.py`, too?

isuruf: There was `tests_require`, but that was deprecated and removed I think. Don't think there's any other option setup.py

inducer: #140

PR #139, opened 2022-09-04
Transition to arraycontext

alexfikl:
Get `sumpy` on array contexts as well!

This eats up the first commit from #118.

TODOs:
- [ ] Pass tests!
- [ ] Port examples!
- [ ] Depends on https://github.com/inducer/boxtree/pull/56

Speculative TODOs:
- [ ] Rip out the timing stuff (not sure, just mentioned by @inducer at some point).
- [ ] Add an array context to do the timing stuff? (likely for a future PR)


sumpy/e2p.py
```diff
@@ -322,15 +315,15 @@ def get_optimized_kernel(self):
                 enforce_variable_access_ordered="no_check")
         return knl
 
-    def __call__(self, queue, **kwargs):
-        knl = self.get_cached_optimized_kernel()
```
isuruf, line 326: Can you keep this line? i.e.
```
knl = self.get_cached_optimized_kernel()
return actx.call_loopy(knl, centers=centers, rscale=rscale, **kwargs)
```
This might be a nitpick, but it saves a few keystrokes everytime I debug these kernels.

sumpy/e2p.py
```diff
@@ -183,16 +174,17 @@ def get_kernel(self):
...
                 name=self.name,
                 assumptions="ntgt_boxes>=1",
                 silenced_warnings="write_race(write_result*)",
-                default_offset=lp.auto,
-                fixed_parameters=dict(dim=self.dim, nresults=len(result_names)),
-                lang_version=MOST_RECENT_LANGUAGE_VERSION)
+                )
 
+        loopy_knl = lp.fix_parameters(loopy_knl,
+            dim=self.dim, nresults=len(result_names))
```
isuruf, line 185: Wouldn't it be better for all the checks that the fixing of parameters happens before the checks happen?
https://github.com/inducer/loopy/blob/a4035dd50189e4b4141ef1e25be094e0e72e0289/loopy/kernel/creation.py#L2522

sumpy/e2p.py
alexfikl: The call is still there in the lines below, just inlined in the `call_loopy`. Did you mean you'd prefer to have the kernel construction on a separate line to be able to break on it?

sumpy/e2p.py
alexfikl: Yeah, definitely. I put this outside because the interface for `make_loopy_program` didn't have it, bu it's probably better to add it to that (loopy is already throwing all sorts of warnings because of this)

Thanks for the push!

alexfikl: Put all of them back in 2332fd5.

sumpy/e2p.py
isuruf: Yes, that's what I mean. Thanks

test/test_fmm.py
```diff
@@ -485,11 +485,7 @@ def test_sumpy_fmm_timing_data_collection(ctx_factory, use_fft, visualize=False)
             fmm_level_to_order=lambda kernel, kernel_args, tree, lev: order)
     from boxtree.fmm import drive_fmm
 
-    timing_data = {}
-    pot, = drive_fmm(actx, wrangler, (weights,), timing_data=timing_data)
-    logger.info("timing_data:\n%s", timing_data)
-
-    assert timing_data
+    pot, = drive_fmm(actx, wrangler, (weights,))
```
isuruf, line 488: Hmm. What's the recommended way to get the timing data?

alexfikl: Not quite sure. I briefly talked about that with @inducer and it should be baked into the array context somehow so it doesn't clutter the wrangler interface.

However, that probably needs a bit of effort to get back into a working state, so not quite sure what a timeline would be..

alexfikl: Just to write out some questions about that to get an idea (from you and @inducer).

* Do you rely on that interface to get results?
* Can this be merged without it?
* Who's going to do the work to put it back? :cry:

isuruf: > Do you rely on that interface to get results?

Yes, I use it to get results for a paper.

> Who's going to do the work to put it back?

I can try if there's an outline of how to reintroduce it. I have not much experience on arraycontexts, so I have no idea what needs to be done.

alexfikl: > I can try if there's an outline of how to reintroduce it. I have not much experience on arraycontexts, so I have no idea what needs to be done.

Yeah, I don't have a good overview either, so that probably needs some discussion. There's a profiling array context here
https://github.com/illinois-ceesd/mirgecom/blob/f9cf0f3557b7d5c89f58e8362ebf50820261219f/mirgecom/profiling.py
but not sure if a similar interface will work for gather timings on the FMM.

> Yes, I use it to get results for a paper.

Ok, then definitely either revert or hold off on merging.

At the moment, the timing code and the manual event handling are not particularly compatible with the `arraycontext` stuff (as it is) and removing both is pretty breaking.

inducer:
Draft merge in #256.

inducer:
Continued in #256, closing here, since I can't seem to force-push here.

PR #143, opened 2022-09-15
Fix for pyrsistent changes

alexfikl:
Latest version seems to have changed some things in an incompatible way.

### DIFF
--- a/sumpy/expansion/diff_op.py
+++ b/sumpy/expansion/diff_op.py
@@ -119,7 +119,8 @@ class LinearPDESystemOperator:
         """
         Returns the total number of dimensions including time
         """
-        return len(self.eqs[0].keys()[0].mi)
+        did = next(iter(self.eqs[0].keys()))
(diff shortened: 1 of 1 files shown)


sumpy/expansion/diff_op.py
```diff
@@ -119,7 +119,8 @@ def total_dims(self):
         """
         Returns the total number of dimensions including time
         """
-        return len(self.eqs[0].keys()[0].mi)
+        dids = list(self.eqs[0].keys())
+        return len(dids[0].mi)
```
isuruf, line 123:
```suggestion
        dids = iter(self.eqs[0].keys())
        return len(next(dids).mi)
```

alexfikl:
The version on `conda-forge` doesn't seem updated yet, so this is probably not going to do anything on the CI.

isuruf (approved):

inducer:
Thx!

PR #146, opened 2022-09-22
Fix src_boxes view

isuruf:
This worked fine before because pypencl didn't care that there was a out-of-bounds access.


inducer:
Thanks! I agree that the patch is right, but I'm still a little lost about the impact. I buy that it didn't make a difference because, to CL, it's just a pointer, and the array bounds aren't respected. Was there an adverse impact from this?

isuruf:
> Was there an adverse impact from this?

No. I noticed this because I was copying `src_boxes` to host to debug and there was an error.

PR #151, opened 2022-11-07
Use a vec array for local_isrc* for coalesced array access

isuruf:
This gives about a 8% speedup

### DIFF
--- a/sumpy/p2p.py
+++ b/sumpy/p2p.py
@@ -664,7 +664,7 @@ class P2PFromCSR(P2PBase):
             # access of 256 bits (assuming double precision) which is
             # optimized for NVIDIA GPUs. On an NVIDIA Titan V, this
             # optimization led to a 8% speedup in the performance.
-            knl = lp.concatenate_memory_layout_of_temporaries(knl,
+            knl = lp.concatenate_arrays(knl,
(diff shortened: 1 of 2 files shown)


inducer:
As discussed: Maybe just transposing the local array can have the same effect.

inducer:
Any chance of turning this into a transform rather than changing the code? In my view, it rather decreases the readability of the code.

sumpy/p2p.py
```diff
@@ -656,6 +656,9 @@ def get_optimized_kernel(self, max_nsources_in_one_box,
             knl = lp.tag_inames(knl, {"itgt_box": "g.0", "inner": "l.0"})
             knl = lp.set_temporary_address_space(knl,
                 ["local_isrc", "local_isrc_strength"], lp.AddressSpace.LOCAL)
+            knl = lp.concatenate_memory_layout_of_temporaries(knl,
```
inducer, line 659: Could you add a comment explaining the conjectured explanation for why this helps performance, the observed increase, qualified with the hardware on which you observed it?

isuruf: Done in becfc2f

inducer:
LGTM, thanks!

PR #156, opened 2023-02-14
Optimize preprocess multipole and postprocess local

isuruf:


inducer:
Since the CIs aren't passing: should this be a draft?

inducer:
LGTM, thanks!

PR #158, opened 2023-02-15
Optimize E2P for GPUs

isuruf:


inducer:
How come the CIs are failing? Is there an unmet dependency somewhere?

isuruf:
Bad merge. Hopefully fixed now.

sumpy/expansion/local.py
```diff
@@ -405,6 +410,14 @@ def loopy_translate_from(self, src_expansion):
             f"A direct loopy kernel for translation from "
             f"{src_expansion} to {self} is not implemented.")
 
+    def get_loopy_evaluator(self, kernels: Sequence[Kernel]) -> lp.TranslationUnit:
```
inducer, line 413: This should be named consistently with the above. How about `loopy_evaluate`?

sumpy/expansion/multipole.py
```diff
@@ -90,17 +90,10 @@ def coefficients_from_source(self, kernel, avec, bvec, rscale, sac=None):
...
         if not self.use_rscale:
             rscale = 1
 
         base_taker = kernel.get_derivative_taker(bvec, rscale, sac)
-        # Following is a no-op, but AxisTargetDerivative.postprocess_at_target and
-        # DirectionalTargetDerivative.postprocess_at_target only handles
-        # DifferentiatedExprDerivativeTaker and sympy expressions, so we need to
-        # make the taker a DifferentitatedExprDerivativeTaker instance.
-        base_taker = DifferentiatedExprDerivativeTaker(base_taker,
-                {tuple([0]*self.dim): 1})
```
inducer, line 103: Could you explain the background to this change?

sumpy/e2p.py
```diff
@@ -81,15 +84,18 @@ def __init__(self, ctx, expansion, kernels,
     def default_name(self):
         pass
 
+    @memoize_method
+    def get_cached_loopy_knl_and_optimizations(self):
+        return self.expansion.get_loopy_evaluator(self.kernels)
```
inducer, line 89: It seems that this would simply fail if E2P were used for M2P, right?

sumpy/expansion/local.py
isuruf: Currently we have a few names that are inconsistent with each other.
```
get_loopy_expansion_formation  (for p2e)
get_loopy_evaluator  (e2p)
loopy_translate_from (m2l)
preprocess_multipole_loopy_knl
postprocess_local_loopy_knl
```
Any suggestions on which to use?

sumpy/e2p.py
```diff
@@ -81,15 +84,18 @@ def __init__(self, ctx, expansion, kernels,
     def default_name(self):
         pass
 
+    @memoize_method
+    def get_cached_loopy_knl_and_optimizations(self):
```
inducer, line 88: I don't like the naming of this, there's already `get_kernel`. The name should reflect that we're talking about the evaluator. Plus whether or not something is memoized isn't typically reflected in its name.

sumpy/expansion/local.py
inducer: I think I would prefer `loopy_purpose_of_thing`. No `get`. But the name *should* indicate whether optimizations are also being returned.

inducer: Btw, I would not be opposed to cleaning up the inconsistencies in this PR.

isuruf: `loopy_evaluate_with_optimizations`?

inducer: That doesn't work: it reads as "evaluate with optimizations", which is not what this is. Maybe making it `loopy_noun_and_optimizations`, so `loopy_evaluator_and_optimizations` might be better?

sumpy/expansion/multipole.py
isuruf: This is not true anymore. `AxisTargetDerivative.postprocess_at_target` and `DirectionalTargetDerivative.postprocess_at_target` now handles `ExprDerivativeTaker`.

sumpy/e2p.py
isuruf: Fixed now.

sumpy/e2p.py
isuruf: No, there's an implementation for M2P as well. (Just not super optimized)

sumpy/kernel.py
```diff
@@ -263,6 +273,18 @@ def get_derivative_coeff_dict_at_source(self, expr_dict):
         """
         return expr_dict
 
+    def get_derivative_coeff_dict_at_target(self, expr_dict):
```
inducer, line 276: - Describe better what the multi-indices on input and output side.
- Generate the source/target docstrings from one source via `format`.
- Name: `apply_target_transformation_to_derivative_coeff_dict`
- Maybe make an explicit type that wraps the `expr_dict` as you suggested.

sumpy/kernel.py
```diff
@@ -263,6 +273,18 @@ def get_derivative_coeff_dict_at_source(self, expr_dict):
...
+        r"""Get the derivative transformation of the expression at target
+        represented by the dictionary expr_dict which is mapping from multi-index
+        `mi` to coefficient `coeff`.
+        Expression represented by the dictionary `expr_dict` is
```
inducer, line 280:
```suggestion
        The expression represented by the dictionary `expr_dict` is
```

sumpy/e2p.py
```diff
@@ -308,30 +366,49 @@ def get_kernel(self):
...
 
         return loopy_knl
 
-    def get_optimized_kernel(self):
-        # FIXME
-        knl = self.get_kernel()
-        knl = lp.tag_inames(knl, {"itgt_box": "g.0"})
+    def get_optimized_kernel(self, max_ntargets_in_one_box):
+        _, optimizations = self.get_loopy_evaluator_and_optimizations()
+        knl = self.get_kernel(max_ntargets_in_one_box=max_ntargets_in_one_box)
```
inducer, line 384: Add a comment to explain the idea?

PR #170, opened 2023-05-21
Make it easier for pocl to generate FMAs

isuruf:


inducer:
This is pretty gross. :slightly_smiling_face: I'm not a huge fan.

https://github.com/pocl/pocl/pull/1227 suggests that you may not have been, either? If so, my vote would be to table this and wait for the pocl change to land. At any rate, thanks for digging into this!

isuruf:
Yeah, not a fan either. I started another PR at https://github.com/inducer/sumpy/pull/172 which is better than this PR IMO.

inducer:
Agreed. Closing this in favor of #172.

PR #176, opened 2023-05-30
Use the fact that helmholtz k is real for better codegen

isuruf:


sumpy/kernel.py
```diff
@@ -519,6 +519,10 @@ def __init__(self, dim, helmholtz_k_name="k",
             scaling = var("I")/4
         elif dim == 3:
             r = pymbolic_real_norm_2(make_sym_vector("d", dim))
+            if allow_evanescent:
+                expr = var("exp")(var("I")*k*r)/r
+            else:
+                expr = (var("cos")(k*r) + var("I")*var("sin")(k*r))/r
             expr = var("exp")(var("I")*k*r)/r
```
alexfikl, line 526:
```suggestion
```
? Seems like `expr` is just overwritten by this line?

isuruf: Oops. Thanks for catching.

inducer:
Thanks!

inducer:
Hmm, whoops. The changes made sense to me, but I missed that the CIs are failing.

sumpy/kernel.py
```diff
@@ -519,7 +519,10 @@ def __init__(self, dim, helmholtz_k_name="k",
             scaling = var("I")/4
         elif dim == 3:
             r = pymbolic_real_norm_2(make_sym_vector("d", dim))
-            expr = var("exp")(var("I")*k*r)/r
+            if allow_evanescent:
+                expr = var("exp")(var("I")*k*r)/r
+            else:
+                expr = var("exp")(var("Ik")*r)/r
```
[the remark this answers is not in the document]
inducer: Add a comment to explain?

sumpy/codegen.py
```diff
@@ -592,6 +592,39 @@ def map_sum(self, expr, *args):
 # }}}
 
 
+# {{{ helmholtz rewrite
+class HelmholtzRewriter(CSECachingIdentityMapper, CallExternalRecMapper):
```
inducer, line 596: Add a docstring to explain intent.

PR #177, opened 2023-06-02
Refactor loopy methods

isuruf:


sumpy/expansion/__init__.py
```diff
@@ -183,7 +183,7 @@ def evaluate(self, kernel, coeffs, bvec, rscale, sac=None):
             in *coeffs*.
         """
 
-    def get_loopy_evaluator(self, kernels: Sequence[Kernel]) -> lp.TranslationUnit:
+    def loopy_evaluator(self, kernels: Sequence[Kernel]) -> lp.TranslationUnit:
```
isuruf, line 186: This would get renamed to `loopy_evaluator_and_optimizations` in #158

inducer:
Thanks! This is definitely a step up from where it was.

PR #179, opened 2023-07-25
Make use of `loopy.TranslationUnit.executor`

inducer:
This avoids long-lived references to CL kernels held by loopy caches.

### DIFF
--- a/sumpy/tools.py
+++ b/sumpy/tools.py
@@ -381,8 +381,17 @@ class OrderedSet(MutableSet):
 
 
 class KernelCacheMixin:
-    @memoize_method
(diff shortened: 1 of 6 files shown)


sumpy/tools.py
```diff
@@ -381,8 +381,17 @@ def __eq__(self, other):
 
 
 class KernelCacheMixin:
+    def get_cached_optimized_kerenl(self, **kwargs):
+        from warnings import warn
+        warn("get_cached_optimized_kerenl is deprecated. "
```
isuruf, line 386:
```suggestion
        warn("get_cached_optimized_kernel is deprecated. "
```

PR #183, opened 2023-08-02
Some small cleanups in sumpy.tools

alexfikl:
Moved these out of #118 to make that nasty diff smaller. This mostly adds docs (the commits should be fairly self-explanatory).


inducer:
Thx!

PR #184, opened 2023-08-04
no_numpy in sumpy kernels

alexfikl:
This takes another chunk from #118. 

Namely, using [make_loopy_program](https://github.com/inducer/arraycontext/blob/main/arraycontext/loopy.py#L42-L64) (or something like it) from `arraycontext` will add a `no_numpy=True` to all kernels. This updates all the tests and other call sites to never pass numpy arrays in.

The stuff in `sumpy.toys` still takes numpy arrays and does all the back and forth behind the scenes.


inducer:
LGTM, thanks!

PR #188, opened 2023-09-23
Pass all arguments to CSE constructor

alexfikl:
Was playing with adding `__slots__` to pymbolic expressions on the `attrs` branch. That doesn't work with default arguments, since it considers them class attributes and `dataclass` gets confused.

Very few expressions had that issue though, mostly just `CommonSubexpression` and `NaN`, and this fixes it here if we ever want to try it!


inducer:
(I know you meant these as just-in-case, but IMO there's no real downside to merging them.)

PR #190, opened 2023-10-08
Fix M2L with preprocessing but no FFT

isuruf:


sumpy/e2e.py
```diff
@@ -332,7 +332,12 @@ def get_translation_loopy_insns(self, result_dtype):
                 [sym.Symbol(f"data{i}")
             for i in range(m2l_translation_classes_dependent_ndata)]
 
-        ncoeff_src = len(self.src_expansion)
+        m2l_translation = self.tgt_expansion.m2l_translation
+        if m2l_translation.use_preprocessing:
+            ncoeff_src = m2l_translation.preprocess_multipole_nexprs(
+                self.tgt_expansion, self.src_expansion)
+        else:
+            ncoeff_src = len(self.src_expansion)
```
inducer, line 340: Maybe `ncoeff_src` should be renamed to clarify that we're talking about "padded" coeffiicents. (Though I'm not sure "padded" is a good adjective.)

inducer, line 340: You mentioned an additional untested code path here.

PR #196, opened 2024-05-12
Enable arraycontext + lazy evaluation + loop fusion

a-alveyblanc:
Initial work toward enabling lazy evaluation and compilation. Builds on #139.

A fair amount of work is left to be done. For example:

- [ ] Tests need to run with `PytatoPyOpenCLArrayContext`
- [ ] Compilation needs to work
    - [ ] Change/remove the use of "views" of arrays
    - [ ] Change in-place updating of arrays (probably in a much more principled way than what's been done so far)
- [ ] Implement transformations, e.g. loop fusion (maybe this is best done in another PR after lazy + compilation are working)
- [ ] Outstanding TODOs from #139 

The list above is not exhaustive. The work so far has been toward enabling `PytatoPyOpenCLArrayContext` and `PyOpenCLArrayContext` to be valid options. However, compilation does not work.


alexfikl:
Having worked on #139.. Yeah, there's still a lot of work to get `sumpy` working nicely with `arraycontext` :cry:

inducer:
Thanks for getting the work on this started. I agree that it's a big job, we'll get it done with time. I suspect we should only keep one of #139 and this. Any thoughts?

alexfikl:
> Thanks for getting the work on this started. I agree that it's a big job, we'll get it done with time. I suspect we should only keep one of #139 and this. Any thoughts?

Maybe we can make a branch with #139 in the main repository (instead of my fork) and we can work on top of that?

a-alveyblanc:
> Maybe we can make a branch with #139 in the main repository (instead of my fork) and we can work on top of that?

I think something like this makes the most sense.

PR #197, opened 2024-05-21
I want to check what goes into a pull request for sumpy

hirish99:
Added a simple class called recurrence that takes in a sumpy PDE and outputs a recurrence relation. However the class is not completely unit-tested or even complete in functionality.

### DIFF
--- a/sumpy/recurrence.py
+++ b/sumpy/recurrence.py
@@ -1,6 +1,6 @@
 __copyright__ = """
-Copyright (C) 2024 Andreas Kloeckner
 Copyright (C) 2024 Hirish Chandrasekaran
+Copyright (C) 2024 Andreas Kloeckner
 """
 
 __license__ = """
@@ -34,16 +34,13 @@ from typing import List
 import sympy as sp
 from sumpy.expansion.diff_op import LinearPDESystemOperator
 from pytools.obj_array import make_obj_array
+from sumpy.symbolic import make_sympy_vec
 
 
 __doc__ = """
 .. autoclass:: Recurrence
-
-.. autofunction:: make_sympy_vec
+.. automodule:: sumpy.recurrence
 """
-
-def make_sympy_vec(name, n):
-    return make_obj_array([sp.Symbol(f"{name}{i}") for i in range(n)])
 
 class Recurrence:
     def __init__(self, sumpy_pde):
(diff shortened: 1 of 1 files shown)


sumpy/recurrence.py
```diff
@@ -0,0 +1,100 @@
+__copyright__ = """
+Copyright (C) 2024 Andreas Kloeckner
+Copyright (C) 2024 Hirish Chandrasekaran
```
inducer, line 3: Your copyright goes first, you're the main author here.

sumpy/recurrence.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+__doc__ = """
+.. autoclass:: Recurrence
+
+.. autofunction:: make_sympy_vec
+"""
+
+def make_sympy_vec(name, n):
+    return make_obj_array([sp.Symbol(f"{name}{i}") for i in range(n)])
```
inducer, line 46: This already exists in `sumpy.symbolic`. Import from there.

sumpy/recurrence.py
```diff
@@ -0,0 +1,100 @@
...
+from itertools import accumulate
+import sumpy.symbolic as sym
+import logging
+from typing import List
+import sympy as sp
+from sumpy.expansion.diff_op import LinearPDESystemOperator
+from pytools.obj_array import make_obj_array
+
+
+__doc__ = """
```
inducer, line 39: Make sure this is included in the main documentation somewhere, via

```
.. automodule:: sumpy.recurrence
```

sumpy/recurrence.py
```diff
@@ -0,0 +1,100 @@
...
+    return make_obj_array([sp.Symbol(f"{name}{i}") for i in range(n)])
+
+class Recurrence:
+    def __init__(self, sumpy_pde):
+        self.sumpy_pde = sumpy_pde
+    
+    '''
+    get_pde_in_recurrence_form
+    Input: 
+        - pde a LinearPDESystemOperator such that assert(len(pde.eqs) == 1) is true.
```
inducer, line 55: Make sure to use proper references here, e.g.
```
:class:`sumpy.expansion.diff_op.LinearSystemPDEOperator`
```

sumpy/recurrence.py
```diff
@@ -0,0 +1,100 @@
...
+__doc__ = """
+.. autoclass:: Recurrence
+
+.. autofunction:: make_sympy_vec
+"""
+
+def make_sympy_vec(name, n):
+    return make_obj_array([sp.Symbol(f"{name}{i}") for i in range(n)])
+
+class Recurrence:
```
inducer, line 48: Is there a good reason this should be a class? (vs. just a function)

sumpy/recurrence.py
hirish99: I don't understand, do I add this line to the __doc__ string?

sumpy/recurrence.py
hirish99: I think so. Right now the code is structured so that you have a sequence of functions where the output of one function goes into the input of another function. There are intermediate representations that I am interested in like the ODE form of the PDE. If I create a recurrence class then I can store all the intermediate representations as attributes and get them immediately if I am interested in them.

PR #210, opened 2024-08-04
Fix ruff issues in notebooks

alexfikl:
ruff 0.5.6 started linting notebooks by default too (under the `preview` flag at least). This fixes all reported issues in the `contrib/translations` notebooks.

### DIFF
--- a/contrib/translations/PDE-reduction and translations.ipynb
+++ b/contrib/translations/PDE-reduction and translations.ipynb
@@ -2,35 +2,39 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 35,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
-    "import pyopencl as cl\n",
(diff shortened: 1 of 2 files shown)


contrib/translations/PDE-reduction and translations.ipynb
```diff
@@ -2,239 +2,185 @@
...
-   "execution_count": 42,
+   "execution_count": null,
    "metadata": {},
-   "outputs": [
-    {
-     "data": {
-      "text/plain": [
-       "<matplotlib.image.AxesImage at 0x7f7f8841ceb8>"
-      ]
-     },
```
alexfikl, line 195: Was some of this output left here on purpose? This clears all the cells in both notebooks.

inducer: No, I think it can go.

inducer:
Thx!

PR #217, opened 2024-10-01
Remove some deprecations from dataclass Expressions

alexfikl:
.. it's just `SpatialConstant`.

The `PersistentHashWalkMapper` changes are probably worth taking a look at, since I'm not quite familiar with that.

### DIFF
--- a/examples/curve-pot.py
+++ b/examples/curve-pot.py
@@ -63,7 +63,7 @@ def draw_pot_figure(aspect_ratio,
     from sumpy.array_context import PyOpenCLArrayContext
     ctx = cl.create_some_context()
     queue = cl.CommandQueue(ctx)
-    actx = PyOpenCLArrayContext(queue, force_device_scalars=True)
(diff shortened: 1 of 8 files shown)


sumpy/expansion/loopy.py
```diff
@@ -104,7 +104,8 @@ def make_e2p_loopy_kernel(
             idx = int(insn.assignee.name[len(result.name):])
             insns[i] = lp.Assignment(
                 assignee=result[idx],
-                expression=result[idx] + insn.expression,
+                # FIXME: this should be removed once loopy and pymbolic agree
+                expression=result[idx] + insn.expression,  # type: ignore[arg-type]
```
alexfikl, line 108: This seems to be because there's some disagreement regarding `ExpressionT` and some scalar types (I think?), so maybe this should wait for loopy changes..

inducer: > so maybe this should wait for loopy changes..

Agreed. I've got those *almost* ready to go.

PR #224, opened 2025-03-17
Fix incorrect scaling of Yukawa kernel

inducer:
Noticed by Shawn Lin.


alexfikl:
There's a test in pytential for the Yukawa-Beltrami operator that would fail with this change (right?)
https://github.com/inducer/pytential/blob/d0321e0681390a4486def0613a59ed7c5c8b3741/test/test_beltrami.py#L131-L140

I haven't looked in this in quite a while, but I'm not seeing why there's a extra `-`, since it seems to match the Laplace version. Maybe worth modifying the comments in there to say why the `-` goes away?

inducer:
@alexfikl Thanks for following up. I agree that this change needs to be backed up better, and I think what I'd actually like is a test for the scaling. What we'd *really* want to test is $\mathcal L G = \delta$, but I don't know how to write a sufficiently general-purpose test for that. Testing a jump relation for the double layer (possibly via QBX) would be another option, though that'd be much easier to do in pytential than sumpy. (Or we could allow the sumpy tests to depend on pytential?) There's [precedent](https://github.com/inducer/pytential/blob/d0321e0681390a4486def0613a59ed7c5c8b3741/test/test_layer_pot.py#L441) for such a test, so this could be a relatively simple copy-paste job.

What do you guys think?

@shawnl00, your input/help would be very welcome here.

alexfikl:
> Testing a jump relation for the double layer (possibly via QBX) would be another option, though that'd be much easier to do in pytential than sumpy. (Or we could allow the sumpy tests to depend on pytential?) There's [precedent](https://github.com/inducer/pytential/blob/d0321e0681390a4486def0613a59ed7c5c8b3741/test/test_layer_pot.py#L441) for such a test, so this could be a relatively simple copy-paste job.

I'm ok with putting more jump tests in pytential for these things. However, from what I recall, the 3D jump tests in there are already quite slow, so it might add quite a bit to the runtime :\

ShawnL00:
> What we'd _really_ want to test is L G = δ , but I don't know how to write a sufficiently general-purpose test for that. Testing a jump relation for the double layer (possibly via QBX) would be another option, though that'd be much easier to do in pytential than sumpy. (Or we could allow the sumpy tests to depend on pytential?) 

Jump test is definitely a good choice. Perhaps we could also explicitly include the leading singularity for $G$ and its related operator $L$. Then, we know what's the expected jump relation.
To make sure we have $L G = \delta$ and not $L G = -\delta$, I do not think analytic derivation / reference is a bad option.

inducer:
> I do not think analytic derivation / reference is a bad option.

I'm more after something that can be verified with code though.

> explicitly include the leading singularity for G and its related operator L

- kernels know their PDE (`get_pde_as_diff_op`)

ShawnL00:
> * How would we check the leading singularity?
I am not sure if this is a good method: to verify the leading singularity, we can define a new kernel by subtracting out the leading singularity, evaluate the double layer along some rays that pass through the boundaries, and then check the continuity conditions on the boundary.

inducer:
I can report at least one bit of happy news: Without this PR, 
```
pycl test_layer_pot.py 'test_3d_jump_relations(_acf, "sp")' 
```
fails in pytential, changed to use the Yukawa kernel. It passes with this PR. So @ShawnL00 is definitely onto something.

PR #225, opened 2025-03-23
Fix use of star expression in index

alexfikl:
Reported by ruff 0.10 (this is a syntax error in Python 3.10).


alexfikl:
Turns out this was a bug in ruff 0.11.2.

PR #227, opened 2025-04-06
Fix ruff 0.11.4 warning for tuple splatting (RUF005)

alexfikl:
Is constructing a list faster than constructing a tuple? Probably doesn't matter..


inducer:
> Is constructing a list faster than constructing a tuple? Probably doesn't matter..

Honestly not sure. Possibly faster? In that spot, I don't think it'll be noticeable.

PR #233, opened 2025-06-18
Add QBMAX expansions

ShawnL00:
I also set CACHING_ENABLED = False in sumpy/tools.py. For different $\tau$ values, do we need to create separate cache files for the kernels?


inducer:
This and #232 look just about identical. I think only one of them is needed?

ShawnL00:
> This and #232 look just about identical. I think only one of them is needed?

[PR #232](https://github.com/inducer/sumpy/pull/232) is a subset of the current pull request, I’ve closed it.

ShawnL00:
@inducer , I made some changes for the LineTaylorExpansion. Let me know if any changes are needed. Thank you!

sumpy/qbx.py
```diff
@@ -309,20 +336,27 @@ def get_kernel(self):
...
-            **kwargs):
+            expansion_vec=None, **kwargs):
         """
         :arg strengths: are required to have area elements and quadrature weights
             already multiplied in.
         """
+        if self._uses_expansion_vec and expansion_vec is None:
+            raise ValueError(
+                "expansion_vec is required when using "
+                "LineTaylorLocalExpansion")
```
ShawnL00, line 347: Pytential tests failed due to the mandatory requirement of `expansion_vec`. One possible fix is:
```python
  if self._uses_expansion_vec and expansion_vec is None:
      expansion_vec = targets - centers
```

I wonder if it is okay.

ShawnL00:
@inducer, this is ready for review now. The pytential API changes are in https://github.com/inducer/pytential/pull/318. I wonder if any modifications are needed.

PR #242, opened 2025-09-09
Improve kernel type annotations

alexfikl:
This is work on top of #240 that adds more types and improves some docs.

I mostly wanted to type the derivative removers for some stuff in `pytential`, but started adding more.. and then realized you're already working on it in #240. Hopefully this isn't interfering too much :worried:

### DIFF
--- a/sumpy/derivative_taker.py
+++ b/sumpy/derivative_taker.py
@@ -39,7 +39,7 @@ __doc__ = """
 """
 
 import logging
-from typing import Any
+from typing import TYPE_CHECKING
 
 import numpy as np
 
@@ -48,6 +48,9 @@ from pytools.tag import tag_dataclass
 import sumpy.symbolic as sym
 from sumpy.tools import add_mi, add_to_sac
 
+
+if TYPE_CHECKING:
+    import sympy as sp
 
 logger = logging.getLogger(__name__)
 
@@ -341,7 +344,7 @@ class HelmholtzDerivativeTaker(RadialDerivativeTaker):
 
 # {{{ DifferentiatedExprDerivativeTaker
 
-DerivativeCoeffDict = dict[tuple[int, ...], Any]
+DerivativeCoeffDict = dict[tuple[int, ...], int | float | complex | sym.Expr]
 
 
 @tag_dataclass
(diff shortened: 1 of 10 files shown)


sumpy/derivative_taker.py
```diff
@@ -341,7 +344,7 @@ def diff(self, mi, q=0):
 
 # {{{ DifferentiatedExprDerivativeTaker
 
-DerivativeCoeffDict = dict[tuple[int, ...], Any]
+DerivativeCoeffDict = dict[tuple[int, ...], int | float | complex | sym.Expr]
```
alexfikl, line 347: Not sure this is correct, but it seems to match most usage?

sumpy/symbolic.py
```diff
@@ -110,6 +110,7 @@ def _find_symbolic_backend():
 # Before adding a function here, make sure it's present in both modules.
 Add = sym.Add
 Basic = sym.Basic
+Expr = sym.Expr
```
alexfikl, line 113: Mostly added this for typing. Not sure it's a good idea?

sumpy/kernel.py
```diff
@@ -427,49 +533,55 @@ def __init__(self, dim: int):
...
+    .. autoattribute:: helmholtz_k_name
+    .. autoattribute:: allow_evanescent
+    """
+
+    mapper_method: ClassVar[str] = "map_helmholtz_kernel"
+
+    helmholtz_k_name: str
+    """The argument name to use for the Helmholtz parameter when generating
+    functions to evaluate this kernel.
     """
```
alexfikl, line 578: Is this actually used anywhere? Couldn't find what it's for..

alexfikl:
This should mostly have the same failures as #240, minus the `pyright` errors, which were rolled into the baseline (as far as I can tell, since I didn't fix anything there :grin:).

alexfikl:
@inducer This looks good to go from my end (only the Stokeslet errors from #240 seem to remain). Let me know if any of the annotations look fishy.

inducer:
Argh. It looks like merging #240 autoclosed this. That was not the intent. Github, are you OK?

PR #245, opened 2025-09-27
Add custom __repr__ for Kernel

alexfikl:
Before:
```python
LaplaceKernel(
	dim=2,
	expression=Call(Variable('log'), (Call(Variable('sqrt'), (Sum((Product((..., ...)), Product((..., ...)))),)),)),
	global_scaling_const=Quotient(1, Product((-2, Variable('pi'))))
)
```
After:
```python
LaplaceKernel(dim=2, expression=log(sqrt(d[0]*d[0] + d[1]*d[1])), global_scaling_const=1 / ((-2)*pi))
```

Thoughts? I mostly added it because the repr for the expression is quite unhelpful in the current version.

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -33594,6 +33594,14 @@
                 }
             },
             {
+                "code": "reportAny",
+                "range": {
+                    "startColumn": 12,
+                    "endColumn": 17,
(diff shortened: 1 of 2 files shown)


inducer:
I agree that seems better.

PR #249, opened 2025-10-19
Update SympyToPymbolic mapper annotations

alexfikl:
https://github.com/inducer/sumpy/actions/runs/18624857397

This fixes some of the new CI failures due to `pymbolic` adding more typing to the interop classes. It also adds a bunch more type annotations to `sumpy.symbolic`, but those are pretty iffy.

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -1214,78 +1214,6 @@
         ],
         "./sumpy/codegen.py": [
             {
-                "code": "reportUnknownParameterType",
-                "range": {
-                    "startColumn": 8,
-                    "endColumn": 21,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportImplicitOverride",
-                "range": {
-                    "startColumn": 8,
-                    "endColumn": 21,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportUnknownParameterType",
-                "range": {
-                    "startColumn": 28,
-                    "endColumn": 32,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportMissingParameterType",
-                "range": {
-                    "startColumn": 28,
-                    "endColumn": 32,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportUnknownArgumentType",
-                "range": {
-                    "startColumn": 21,
-                    "endColumn": 25,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportUnknownMemberType",
-                "range": {
-                    "startColumn": 24,
-                    "endColumn": 63,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportUnknownArgumentType",
-                "range": {
-                    "startColumn": 70,
-                    "endColumn": 74,
-                    "lineCount": 1
-                }
-            },
-            {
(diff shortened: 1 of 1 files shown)


sumpy/symbolic.py
```diff
@@ -153,48 +160,51 @@ def UnevaluatedExpr(x):  # noqa: N802
...
+        return Pow(a, b)
 else:
-    def doit(expr):
+    def doit(expr: Expr) -> Expr:
         return expr.doit()
 
-    def unevaluated_pow(a, b):
-        return sym.Pow(a, b, evaluate=False)
+    def unevaluated_pow(a: Expr, b: complex | Expr) -> ArithmeticExpr:
+        return Pow(a, b, evaluate=False)
```
alexfikl: This is most likely incorrect, but it seemed to be what `sympy.Pow` wanted? Not sure..

sumpy/symbolic.py
```diff
@@ -245,20 +255,21 @@ def checked_cse(exprs, symbols=None):
...
+def sym_real_norm_2(x: Matrix) -> Expr:
     return sym.sqrt((x.T*x)[0, 0])
 
 
-def pymbolic_real_norm_2(x):
-    from pymbolic import var
-    return var("sqrt")(np.dot(x, x))
+def pymbolic_real_norm_2(
+            x: ObjectArray1D[ArithmeticExpression]) -> ArithmeticExpression:
+    return prim.Variable("sqrt")(x.dot(x))
```
alexfikl: Worth adding `ObjectArray.dot` to pytools?

sumpy/symbolic.py
```diff
@@ -268,12 +279,14 @@ def vector_xreplace(expr, from_vec, to_vec):
...
-def find_power_of(base, prod):
+def find_power_of(base: sym.Basic, prod: sym.Basic) -> int | sym.Basic:
     remdr = sym.Wild("remdr")
     power = sym.Wild("power")
+
     result = prod.match(remdr*base**power)
     if result is None:
         return 0
+
     return result[power]
```
alexfikl: This doesn't seem used and it also doesn't work with `symengine` as far as I can tell?

sumpy/symbolic.py
```diff
@@ -294,92 +307,103 @@ class SpatialConstant(prim.Variable):
...
+            if function.name == "hankel_1":
+                args = [self.rec(param) for param in expr.parameters]
+                args.append(sym.sympify(0))
+                return Hankel1(*args)
+            elif function.name == "bessel_j":
+                args = [self.rec(param) for param in expr.parameters]
+                args.append(sym.sympify(0))
+                return BesselJ(*args)
+
+        return PymbolicToSympyMapper.map_call(self, expr)
```
alexfikl: Is this correct? pyright was complaining that `expr.function.name` does not have  `name` because `function: Expression`, but it wasn't clear how they're constructed here. I'm guessing the CI will fail if this is wrong 😁

sumpy/symbolic.py
inducer: What's incorrect about it? I'm not sure I'm seeing anything wrong.

sumpy/symbolic.py
inducer: LGTM.

inducer:
Ty!

PR #250, opened 2025-10-22
Expn typing

inducer:
- [x] Needs https://github.com/inducer/pymbolic/pull/215

cc @alexfikl for awareness

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -606,208 +606,16 @@
             {
                 "code": "reportUnknownParameterType",
                 "range": {
-                    "startColumn": 23,
-                    "endColumn": 34,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportMissingParameterType",
-                "range": {
-                    "startColumn": 23,
-                    "endColumn": 34,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportUnknownMemberType",
-                "range": {
-                    "startColumn": 8,
-                    "endColumn": 24,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportUnannotatedClassAttribute",
-                "range": {
-                    "startColumn": 13,
-                    "endColumn": 24,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportUnknownMemberType",
-                "range": {
-                    "startColumn": 8,
-                    "endColumn": 33,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportUnannotatedClassAttribute",
-                "range": {
-                    "startColumn": 13,
-                    "endColumn": 33,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportUnknownMemberType",
-                "range": {
-                    "startColumn": 56,
-                    "endColumn": 73,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportUnannotatedClassAttribute",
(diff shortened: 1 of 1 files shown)


doc/conf.py
```diff
@@ -39,6 +39,9 @@
     "obj_array.ObjectArray1D": "obj:pytools.obj_array.ObjectArray1D",
     # sympy
     "sp.Matrix": "class:sympy.matrices.dense.DenseMatrix",
+    "sym.Expr": "class:sympy.core.expr.Expr",
+    "sym.Symbol": "class:sympy.core.symbol.Symbol",
+    "sym.Matrix": "class:sympy.matrices.dense.DenseMatrix",
```
alexfikl: Hm, it just occurred to me that `sym.Matrix` is not an `sp.Expr` or an `sp.Basic`, but I don't think we want to handle those in `pymbolic`, right?

sumpy/test/test_misc.py
```diff
@@ -592,9 +592,9 @@ def get_pde_as_diff_op(self):
 def test_get_storage_index(order, knl, compressed):
     dim = knl.dim
     if compressed:
-        wrangler = LinearPDEBasedExpansionTermsWrangler(order, dim, knl=knl)
+        wrangler = LinearPDEBasedExpansionTermsWrangler(order, dim, None, knl=knl)
     else:
-        wrangler = FullExpansionTermsWrangler(order, dim)
+        wrangler = FullExpansionTermsWrangler(order, dim, None)
```
alexfikl:
```suggestion
        wrangler = FullExpansionTermsWrangler(order, dim, max_mi=None)
```
? Reads a bit better easier for those of us who are not as familiar with the interface :grin:

EDIT: Hm, this is in a test, so maybe nevermind..

sumpy/symbolic.py
```diff
@@ -332,7 +332,7 @@ def map_Mul(self, expr: Mul) -> Expression:
 
 class PymbolicToSympyMapperWithSymbols(PymbolicToSympyMapper):
     @override
-    def map_variable(self, expr: prim.Variable) -> Basic:
+    def map_variable(self, expr: prim.Variable) -> sym.Basic:
```
alexfikl:
```suggestion
    def map_variable(self, expr: prim.Variable) -> Basic:
```
? :grin:

sumpy/symbolic.py
```diff
@@ -163,13 +163,13 @@ def UnevaluatedExpr(x):  # noqa: N802
     def doit(expr: Expr) -> Expr:
         return expr
 
-    def unevaluated_pow(a: Expr, b: complex | Expr) -> ArithmeticExpr:
```
alexfikl: I don't think `ArithmeticExpr` is used anywhere else, so it can probably go? I just added it for this because sympy is nice enough to evaluate `Pow(2, 3)`, but that's not actually the case here..

sumpy/kernel.py
```diff
@@ -262,7 +262,7 @@ def get_code_transformer(self) -> Callable[[Expression], Expression]:
         return lambda expr: expr
 
     @abstractmethod
-    def get_expression(self, dist_vec: sp.Matrix) -> ArithmeticExpr:
```
alexfikl: `ArithmeticExpr` can probably go here too? There are different aliases for it in `kernel` and `symbolic` too, which is not great :(

sumpy/symbolic.py
inducer: Done.

PR #252, opened 2025-11-08
Improve typing in codegen

alexfikl:
Also 
* Adds a bunch more types to expansions (mostly return values and fixes some errors within various methods).
* Removes `CallExternalRecMapper` and the combine mappers machinery, as discussed.

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -678,1138 +678,18 @@
                 }
             },
             {
-                "code": "reportImplicitOverride",
-                "range": {
-                    "startColumn": 8,
-                    "endColumn": 18,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportMissingParameterType",
-                "range": {
-                    "startColumn": 25,
-                    "endColumn": 40,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportMissingParameterType",
-                "range": {
-                    "startColumn": 42,
-                    "endColumn": 54,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportOperatorIssue",
-                "range": {
-                    "startColumn": 19,
-                    "endColumn": 33,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportOperatorIssue",
-                "range": {
-                    "startColumn": 26,
-                    "endColumn": 32,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportIncompatibleMethodOverride",
-                "range": {
-                    "startColumn": 8,
-                    "endColumn": 26,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportImplicitOverride",
-                "range": {
-                    "startColumn": 8,
-                    "endColumn": 26,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportUnknownParameterType",
-                "range": {
-                    "startColumn": 33,
-                    "endColumn": 39,
-                    "lineCount": 1
-                }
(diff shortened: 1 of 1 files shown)


sumpy/codegen.py
```diff
@@ -124,10 +134,17 @@ def not_supported(self, expr: object) -> Expression:
...
+    @override
+    def with_types(self,
+                   arg_id_to_dtype: Mapping[int | str, LoopyType],
+                   clbl_inf_ctx: CallablesInferenceContext,
+               ) -> tuple[BesselJvvp1, CallablesInferenceContext]:
         from loopy.types import NumpyType
 
         for i in arg_id_to_dtype:
+            if isinstance(i, str):
+                raise TypeError(f"{self.name} cannot handle keyword arguments")
```
alexfikl: I assumed this is what a string means in there? Is that right?

sumpy/expansion/local.py
```diff
@@ -417,21 +410,28 @@ def translate_from(self,
             # Use the axis as the first dimension to vary so that the below
             # algorithm is O(p^{d+1}) for full and O(p^{d}) for compressed
             dims = [axis, *list(range(axis)), *list(range(axis + 1, self.dim))]
+
             # Start with source coefficients. Gets updated after each axis.
-            cur_dim_input_coeffs = src_coeffs
+            cur_dim_input_coeffs = cur_dim_output_coeffs = list(src_coeffs)
```
alexfikl: This just gives a default value to `cur_dim_output_coeffs` because `pyright` was complaining about it. Is it confusing?

sumpy/expansion/local.py
```diff
@@ -515,25 +521,21 @@ def m2l_translation(self) -> M2LTranslationBase:
...
             from sumpy.expansion.m2l import DefaultM2LTranslationClassFactory
+
             factory = DefaultM2LTranslationClassFactory()
-            return (
-                factory.get_m2l_translation_class(self.kernel, self.__class__)())
-            from sumpy.expansion.m2l import DefaultM2LTranslationClassFactory
-            factory = DefaultM2LTranslationClassFactory()
-            return factory.get_m2l_translation_class(self.kernel,
-                self.__class__)()
+            return factory.get_m2l_translation_class(self.kernel, self.__class__)()
```
alexfikl: This seems to have been a merge gone wrong? Didn't check to see where it comes from..

sumpy/expansion/local.py
inducer: It is confusing to me. Would it instead be possible to manufacture a list of bogus values?

sumpy/expansion/local.py
```diff
@@ -506,13 +521,13 @@ class LinearPDEConformingVolumeTaylorLocalExpansion(
...
-    m2l_translation_override: M2LTranslationBase | None = field(
-                                                kw_only=True, default=None)
+    m2l_translation_override: M2LTranslationBase | None = (
+            field(kw_only=True, default=None))
 
     @property
     @abstractmethod
     def mpole_expn_class(self) -> type[HankelBased2DMultipoleExpansion]:
-        ...
+        pass
```
inducer: I think `...` is better for the body of abstract methods, since it's visually distinct from an empty method.

sumpy/expansion/__init__.py
```diff
@@ -142,7 +142,7 @@ def get_source_args(self) -> Sequence[KernelArgument]:
 
     @abstractmethod
     def get_storage_index(self, mi: MultiIndex) -> int:
-        ...
+        pass
```
inducer: As elsewhere.

inducer (commented):
Thx!

PR #256, opened 2025-12-04
Towards array context (2025 merge)

inducer:
Resumed from #139.

----

Relevant parts of summary from #139 follow:

### Remaining porting issues

- [ ] Move caching to the array context from `KernelCacheMixin`. (now #257)
- [ ] Handle kernel optimizations. (now #258(
	- Currently it uses `get_cached_executor()`, which branches on `OPT_ENABLED`, but we need to move to `actx.call_loopy`.
	- Move to `transform_loopy_program`? Would need to also add some fancy tags or something to customize it?
- [ ] Need to move e.g. `register_optimization_preambles` to the array context somehow too. (now also #258)
	- The kernels no longer have a context or device, so they can't do that.
	- Probably need to move this to `actx.transform_loopy_program` somehow?

### Upstream / downstream PRs

- [x] Depends on https://github.com/inducer/boxtree/pull/56
- [x] Depends on https://github.com/inducer/loopy/pull/677
- [ ] Depends on: https://github.com/inducer/pytential/pull/179

### Speculative TODOs

- [x] Should some classes clone the `actx`? Here e.g. `SumpyTreeIndependentDataForWrangler` but also in `boxtree`.
	- The big worry is that cloning throws away any caches on the `actx`.
	- Pretty much all `*Builder` classes can be functions + all other memoized kernels can be functions, but should leave that for later since it can be done without breaking the world.
- [x] Rip out the timing stuff (not sure, just mentioned by @inducer at some point).
	- This would also need to be done in `boxtree`?
- [ ] Add an array context to do the timing stuff? (now #259)


inducer:
FYI: As in boxtree, I've moved towards annotating with `ArrayContext` instead of `PyOpenCLArrayContext`. I like that direction because it conveys the ambition (*cough*) that all this will work with any actx. It also prevents having to splatter actx-type asserts everywhere downstream. If you hate it, it's a single commit that could be unwound.

alexfikl:
> FYI2: The local tests are now passing here, so maybe this can become #139? (I.e. I would push this code there?) Does that sound OK?

Definitely! I'm fine with continuing the work here too, whatever's more convenient.

alexfikl:
> Do you still have those? Could you reapply them?

Done! :grin:

PR #260, opened 2026-01-20
l2l coeffs test

ShawnL00:
l2l coeffs test

PR #272, opened 2026-03-18
Verify translation difference formulas

inducer:
Subsumes #260 and #261.

### DIFF
--- a/sumpy/test/coeff_test_tools.py
+++ b/sumpy/test/coeff_test_tools.py
@@ -1,0 +1,68 @@
+from __future__ import annotations
+
+
+__copyright__ = """
+Copyright (C) 2026 Shawn/Chaoqi Lin
+"""
+
+__license__ = """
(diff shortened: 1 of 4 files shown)


sumpy/test/coeff_test_tools.py
```diff
@@ -0,0 +1,68 @@
+from __future__ import annotations
+
+
+__copyright__ = """
+Copyright (C) 2026 Shawn/Chaoqi Lin
+"""
+
+__license__ = """
```
inducer, line 8: @ShawnL00 Please let me know if you agree with this licensing, across this whole PR.

inducer:
Good to go from my end, subject to license approval.

sumpy/test/coeff_test_tools.py
ShawnL00: Yes, thank you!

inducer: Thanks! In it goes.

PR #279, opened 2026-05-08
Implement kernel rewriting

alexfikl:
This implements rewriting a target kernel (e.g. Stokeslet) in terms of a base kernel (e.g. biharmonic) as a linear constant-coefficient combination of derivatives.

Currently, it just hoists `get_deriv_relation_kernel` from https://github.com/inducer/pytential/pull/162 with some minimal changes:
* Clean up the retry logic for the LU factorization.
* Add a loop on the max order of derivatives.

TODO:
* [ ] Make sure the LU based decomposition works the same as before.
* [ ] Add a bunch of tests for it (based on the tests in `pytential`).
* [ ] Implement a hopefully better (but likely slower?) purely-symbolic version.
* [ ] Hook into documentation.

cc @isuruf

### DIFF
--- a/sumpy/kernel_rewrite.py
+++ b/sumpy/kernel_rewrite.py
@@ -1,0 +1,352 @@
+from __future__ import annotations
+
+
+__copyright__ = """
+Copyright (C) 2012 Andreas Kloeckner
+Copyright (C) 2020 Isuru Fernando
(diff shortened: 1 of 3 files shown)


sumpy.html
inducer, line 1: 👋

PR #281, opened 2026-05-09
Separate ElasticityKernel and Stokeslet

alexfikl:
This makes the `StokesletKernel` an `ExpressionKernel` instead of inheriting from `ElasticityKernel`. Besides that, it also
1. Removes the `__new__` implementations. This is breaking (since `ElasticityKernel` no longer returns a Stokeslet in special cases).
2. ~Deprecates passing a non-string value for `viscosity_mu` and `poisson_ratio` to match the existing scalar kernels.~
2. Renames `viscosity_mu` and `poisson_ratio` to `viscosity_mu_name` and `poisson_ratio_name`.
3. Makes it an error to pass non-string values to these things. This matches all the other kernels.

### DIFF
--- a/sumpy/kernel.py
+++ b/sumpy/kernel.py
@@ -861,6 +861,11 @@ class ElasticityKernel(ExpressionKernel):
                  "raise a TypeError starting with Q1 2027.",
                  DeprecationWarning, stacklevel=2)
 
+            if poisson_ratio == 0.5:  # noqa: RUF069
+                raise ValueError(
+                    f"{type(self)} cannot accept 'poisson_ratio' of 0.5 "
+                    "(use StokesletKernel instead)")
+
             nu = poisson_ratio
 
         d = make_sym_vector("d", dim)
@@ -893,9 +898,19 @@ class ElasticityKernel(ExpressionKernel):
 
     @override
     def __reduce__(self):
+        # TODO: remove this once we stop allowing non-str values for viscosity_mu
+        viscosity_mu = (
+            self.viscosity_mu.name
+            if isinstance(self.viscosity_mu, SpatialConstant)
+            else self.viscosity_mu)
+        poisson_ratio = (
+            self.poisson_ratio.name
+            if isinstance(self.poisson_ratio, SpatialConstant)
+            else self.poisson_ratio)
+
         return (
             type(self),
-            (self.dim, self.icomp, self.jcomp, self.viscosity_mu, self.poisson_ratio))
+            (self.dim, self.icomp, self.jcomp, viscosity_mu, poisson_ratio))
 
     @property
     @override
@@ -979,6 +994,9 @@ class StokesletKernel(ExpressionKernel):
                  "argument will be removed in Q1 2027.",
                  DeprecationWarning, stacklevel=2)
 
+            if poisson_ratio != 0.5:  # noqa: RUF069
+                raise ValueError("'poisson_ratio' must be 0.5 (if provided)")
+
         if isinstance(viscosity_mu, str):
             mu = SpatialConstant(viscosity_mu)
         else:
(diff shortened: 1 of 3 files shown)


sumpy/kernel.py
```diff
@@ -483,6 +496,13 @@ def is_complex_valued(self) -> bool:
 # {{{ PDE kernels
 
 class LaplaceKernel(ExpressionKernel):
+    r"""A kernel for the Laplace equation (see e.g. Theorem 6.2 from [Kress2013]_).
+
+    .. math::
+
+        \Delta K(\mathbf{x}, \mathbf{y}) = \delta(\mathbf{x} - \mathbf{y}).
```
alexfikl, line 508: I think this is right, i.e. it's not $-\Delta K = \delta$, but pointing it out just to make sure. :grin:

sumpy/kernel.py
```diff
@@ -779,59 +826,49 @@ class ElasticityKernel(ExpressionKernel):
...
+                 icomp: int,
+                 jcomp: int,
                  viscosity_mu: float | str | SpatialConstant = "mu",
                  poisson_ratio: float | str | SpatialConstant = "nu") -> None:
         if isinstance(viscosity_mu, str):
             mu = SpatialConstant(viscosity_mu)
         else:
+            warn("Passing a non-str 'viscosity_mu' is deprecated. This will "
+                 "raise a ValueError starting with Q1 2027.",
+                 DeprecationWarning, stacklevel=2)
```
alexfikl: Ideally these would be just `viscosity_mu_name` to match Helmholtz and Yukawa?

sumpy/kernel.py
```diff
@@ -892,48 +929,126 @@ def get_pde_as_diff_op(self) -> LinearPDESystemOperator:
...
+    @property
+    @override
+    def is_complex_valued(self) -> bool:
+        return False
+
+    @memoize_method
+    @override
+    def get_args(self) -> Sequence[KernelArgument]:
+        # TODO: remove this once we stop allowing non-str values for viscosity_mu
+        if isinstance(self.viscosity_mu, SpatialConstant):
```
alexfikl: In `ElasticityKernel`, this uses `sumpy.tools.get_all_variables`, but it didn't see worth going through a `DependencyMapper`, since we know it's just a `SpatialConstant`. Should I change it there too?

sumpy/kernel.py
```diff
@@ -892,48 +929,126 @@ def get_pde_as_diff_op(self) -> LinearPDESystemOperator:
...
-                f"got '{poisson_ratio}'")
+                 dim: int,
+                 icomp: int,
+                 jcomp: int,
+                 viscosity_mu: float | str | SpatialConstant = "mu",
+                 poisson_ratio: float | str | SpatialConstant | None = None) -> None:
+        if poisson_ratio is not None:
+            warn(f"Passing 'poisson_ratio' to {type(self)} is deprecated. This "
+                 "argument will be removed in Q1 2027.",
+                 DeprecationWarning, stacklevel=2)
```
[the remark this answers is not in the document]
alexfikl: Fair enough. Fixed.

sumpy/kernel.py
```diff
@@ -779,59 +826,49 @@ class ElasticityKernel(ExpressionKernel):
...
             mu = viscosity_mu
 
         if isinstance(poisson_ratio, str):
             nu = SpatialConstant(poisson_ratio)
         else:
+            warn("Passing a non-str 'poisson_ratio' is deprecated. This will "
+                 "raise a ValueError starting with Q1 2027.",
+                 DeprecationWarning, stacklevel=2)
+
             nu = poisson_ratio
```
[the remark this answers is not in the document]
alexfikl: Will go with pickling a string/float for now.

alexfikl:
@inducer This should also be good for a look (the CI passed once). It doesn't seem to break `pytential`, so not bad for all the breaking changes it makes :grin:

inducer:
Yep! Much easier on the eyes now without all that compatibility glop.

inducer:
Thanks!

PR #282, opened 2026-05-09
Add the Brinkman kernel

alexfikl:
This adds the Brinkman kernel and the corresponding Brinkman stress tensor. I'm not particularly sure the scaling is right (the 2-3 papers I've found all had different scaling), but I've checked (in Mathematica) that
* The kernel itself (with the pressure kernel) satisfies the Brinkman equation.
* The stress matches $-p_j \delta_{ik} + \mu (\partial_i K_{jk} + \partial_k K_{ij})$.
* The limits as $k \to 0$ gives back the Stokeslet and Stresslet with the scalings that we already have.

Also added these to some tests: `test_pde_check_kernels` and `test_translations`. There's probably more things we could test, but it should be ok for now.

TODO:
* [x] There's also a stress tensor for this (matching the Stresslet).

### DIFF
--- a/sumpy/kernel.py
+++ b/sumpy/kernel.py
@@ -106,6 +106,12 @@ PDE kernels
 .. autoclass:: LineOfCompressionKernel
     :show-inheritance:
     :members: mapper_method
+.. autoclass:: BrinkmanletKernel
+    :show-inheritance:
+    :members: mapper_method
+.. autoclass:: BrinkmanStressKernel
+    :show-inheritance:
+    :members: mapper_method
 
 Derivatives
 -----------
@@ -724,7 +730,7 @@ class YukawaKernel(ExpressionKernel):
     @property
     @override
     def is_complex_valued(self) -> bool:
-        # FIXME
+        # FIXME: 2D uses Hankel functions (complex-valued); 3D uses real exp
         return True
 
     @override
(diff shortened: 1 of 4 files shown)


sumpy/kernel.py
```diff
@@ -1119,6 +1122,124 @@ def get_pde_as_diff_op(self) -> LinearPDESystemOperator:
...
+
+        if dim == 2:
+            # transforming Bessel functions to Hankel functions using [2]
+            K0 = var("pi") * var("I") / 2 * var("hankel_1")(0, var("I") * R)  # noqa: N806
+            K1 = -var("pi") / 2 * var("hankel_1")(1, var("I") * R)            # noqa: N806
+
+            # [1] Equations 7.7.5 and 7.7.6
+            a = K0 + K1 / R - 1 / R**2
+            b = 2 / R**2 - 2 * K1 / R - K0
+            expr = a * delta_ij + b * d[icomp] * d[jcomp] / r ** 2
```
alexfikl, line 1191: Does this `a * delta_ij` get flattened somewhere? Is it better to flatten it here?

alexfikl:
For what it's worth, doing the `test_translations` tests for these Brinkman things is quite slow compared to the other kernels in there :grin:

sumpy/kernel.py
inducer: Yep! Although that's surprisingly recent. It got noticed by @ShawnL00  since some flop counts were off. 🙂 

https://github.com/inducer/sumpy/commit/a1bf1e2352ae518b8040bfc9cfc6b3f8993b11cd

inducer:
> For what it's worth, doing the `test_translations` tests for these Brinkman things is quite slow compared to the other kernels in there 😁

That's fine! Hopefully we can get them fast with recurrence-based code generation. (cc @hirish99)

inducer:
Thanks!

sumpy/kernel.py
alexfikl: https://github.com/inducer/sumpy/pull/283 :grin:

PR #285, opened 2026-05-22
Make Fourier symbol from a LinearPDESystemOperator

alexfikl:
Some of this functionality was already in there in `as_scalar_pde` (the "make a polynomial out of this PDE" part), but this introduces a separate Fourier function.


alexfikl:
@inducer This should be good for a look now. I've fixed the Copilot comments and renamed `dim` to `spatial_dim`.

inducer:
Looks great, thanks!

PR #286, opened 2026-05-29
Remove some unnecessary uses of scipy

alexfikl:
1. Replaced the uses of `scipy.special.factorial` with `mi_factorial` in coefficient tests. Hopefully fixes https://gitlab.tiker.net/inducer/sumpy/-/jobs/957466.
2. Replaced uses of `scipy.fftpack` with `numpy.fft`.
3. Replaced some uses of Legendre and Chebyshev functions in `point_calculus`.

### DIFF
--- a/doc/misc.rst
+++ b/doc/misc.rst
@@ -111,7 +111,7 @@ The FAQ is maintained collaboratively on the
 Acknowledgments
 ===============
 
-Work on meshmode was supported in part by
+Work on sumpy was supported in part by
 
 * the US National Science Foundation under grant numbers DMS-1418961,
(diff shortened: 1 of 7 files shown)


doc/misc.rst
```diff
@@ -111,7 +111,7 @@ The FAQ is maintained collaboratively on the
 Acknowledgments
 ===============
 
-Work on meshmode was supported in part by
+Work on sumpy was supported in part by
```
alexfikl, line 114: ?

inducer: Indeed, thanks! Also, needs updating for new grant numbers. I'll put in on the list.

inducer:
Thanks!

PR #287, opened 2026-06-02
Add a SystemKernel wrapper

alexfikl:
This is very drafty at the moment.

However, the point is to:
1. Rename `Kernel` to `ScalarKernel`
2. Introduce a `SystemKernel` for things like Stokes. Currently, the `SystemKernel` is intentionally more barebones than a `ScalarKernel` because it cannot (and maybe will not?) be used for codegen.
3. Implement and test the various kernels we have as `SystemKernel`.

### DIFF
--- a/sumpy/kernel.py
+++ b/sumpy/kernel.py
@@ -45,7 +45,7 @@ import pymbolic.primitives as prim
 from pymbolic import Expression, var
 from pymbolic.mapper import CSECachingMapperMixin, IdentityMapper
 from pymbolic.primitives import make_sym_vector
-from pytools import memoize_method
+from pytools import keyed_memoize_method, memoize_method, obj_array
 
 import sumpy.symbolic as sym
 from sumpy.derivative_taker import (
@@ -70,6 +70,8 @@ Kernel interface
 
 .. autoclass:: KernelArgument
 .. autoclass:: ScalarKernel
+    :show-inheritance:
+.. autoclass:: SystemKernel
     :show-inheritance:
 
 Symbolic kernels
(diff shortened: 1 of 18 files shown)


sumpy/kernel.py
```diff
@@ -198,8 +243,25 @@ def name(self) -> str:
...
+
+    return expr
+
+
+KernelObjectT = TypeVar("KernelObjectT")
+
+
 @dataclass(frozen=True)
-class ScalarKernel(ABC):
+class GenericKernel(ABC):
```
inducer: Nah.

sumpy/kernel.py
```diff
@@ -513,6 +598,47 @@ def is_complex_valued(self) -> bool:
 one_kernel_3d = OneKernel(3)
 
 
+@dataclass(frozen=True, repr=False)
+class ExpressionSystemKernel(SystemKernel, ABC):
```
inducer: Nah.

sumpy/kernel.py
```diff
@@ -249,6 +298,55 @@ def __repr__(self) -> str:
...
+    def get_pde_as_diff_op(self) -> LinearPDESystemOperator:
+        r"""
+        :returns: the PDE for the kernel as a
+            :class:`sumpy.expansion.diff_op.LinearPDESystemOperator` object
+            :math:`\mathcal{L}`, where :math:`\mathcal{L}(u) = 0` is the PDE.
+        """
+
+
+@dataclass(frozen=True, repr=False)
+class ScalarKernel(GenericKernel, ABC):
```
inducer:
```
sudo get-me-a-system-kernel-and-index -f
```

sumpy/kernel.py
```diff
@@ -1547,6 +1672,351 @@ def get_pde_as_diff_op(self):
...
+
+    def __init__(self,
+                 dim: int,
+                 viscosity_mu_name: str = "mu",
+                 poisson_ratio_name: str = "nu") -> None:
+        components = {}
+
+        for i in range(dim):
+            for j in range(i, dim):
+                components[i, j] = components[j, i] = ElasticityKernel(
```
inducer: - Deprecate `ElasticityKernel`.
- Rename to `ElasticityComponentKernel`.

?

sumpy/kernel.py
alexfikl: Done!

sumpy/kernel.py
alexfikl: Done!

alexfikl:
The `optype<0.18` on the CI seems to be stopping installing `numpy==2.5.0`, which is causing the mismatch between my local baseline and the CI. A bit annoying..

PR #292, opened 2026-06-27
Add the stress kernel for linear elasticity

alexfikl:
1. This adds a little base class for every PDE (Stokes, elasticity, Brinkman ATM) to avoid some repetition.
2. Adds the elasticity stress tensor. This is essentially just
```math
\Sigma_{ijk} = \lambda \left( \frac{\partial G_{lk}}{\partial x_l} \right) \delta_{ij} + \mu \left( \frac{\partial G_{ik}}{\partial x_j} + \frac{\partial G_{jk}}{\partial x_i} \right)
```
like for Stokes and Brinkman, but written out explicitly. https://github.com/inducer/pytential/pull/162 has it written in terms of the Laplace kernel and the Stresslet kernel, since it's just
```math
\Sigma_{ijk}(\mathbf{r}) = -\frac{1}{8\pi(1-\nu)r^2} \left[ (1-2\nu)\left( \delta_{jk}\frac{r_i}{r} + \delta_{ik}\frac{r_j}{r} - \delta_{ij}\frac{r_k}{r} \right) + 3\frac{r_i r_j r_k}{r^3} \right]
```
where the first 3 terms are Laplace derivatives and the last term is exactly the Stresslet. Since at the end of the day we want to rewrite all of this in terms of the biharmonic, I figure it shouldn't matter.

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -9242,14 +9242,6 @@
                 }
             },
             {
-                "code": "reportArgumentType",
-                "range": {
-                    "startColumn": 25,
-                    "endColumn": 48,
-                    "lineCount": 1
-                }
-            },
-            {
                 "code": "reportReturnType",
                 "range": {
                     "startColumn": 15,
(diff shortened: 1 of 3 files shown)


alexfikl:
This is the code for it in https://github.com/inducer/pytential/pull/162:
https://github.com/isuruf/pytential/blob/ba5cc19daa3c921308bb7272f5786961cf11fd71/pytential/symbolic/elasticity.py#L432-L462

inducer:
Thx!

PR #295, opened 2026-07-17
Switch to ruff:ignore human-readable rules

alexfikl:
Just ran `ruff check --fix` :grin: The baseline update seems to be necessary because the autofix broke some imports into multiple lines.


inducer:
Thx!

PR #300, opened 2026-08-19
Add type annotations around sumpy.fmm

alexfikl:
xref: https://github.com/inducer/boxtree/pull/125


sumpy/__init__.py
```diff
@@ -72,15 +73,13 @@
 
 # {{{ optimization control
 
-OPT_ENABLED = True
-
 OPT_ENABLED = "SUMPY_NO_OPT" not in os.environ
```
alexfikl, line 76: I assume this was just a delete gone wrong?

sumpy/__init__.py
```diff
@@ -91,36 +90,55 @@ def set_optimization_enabled(flag):
...
+                 new_flag: bool,
+                 new_no_cache_kernels: tuple[str, ...] = ()) -> None:
         self.new_flag = new_flag
+        self.previous_flag = None
         self.new_no_cache_kernels = new_no_cache_kernels
+        self.previous_no_cache_kernels = None
+
+    def __enter__(self) -> None:
+        if self.previous_flag is not None or self.previous_no_cache_kernels is not None:
+            raise RuntimeError("cannot reuse the 'CacheMode' context manager")
```
alexfikl, line 126: `pyright` was complaining about setting attributes outside of `__init__`, so I reworked this a bit. It now yells if it's nested like `with ... as cm: with cm: ..`.

sumpy/tools.py
```diff
@@ -49,14 +48,16 @@
 
 
 if TYPE_CHECKING:
-    import numpy
     from numpy.typing import DTypeLike
     from optype.numpy import Array2D
+    from pyvkfft.opencl import VkFFTApp
```
alexfikl, line 53: Is it ok to make this a dependency for typing? It's mostly just used in some return types.

sumpy/tools.py
```diff
@@ -276,19 +277,28 @@ class ScalingAssignmentTag(Tag):
...
+
     def __init__(self,
-            target_kernels: list[ScalarKernel],
-            source_kernels: list[ScalarKernel],
-            strength_usage: list[int] | None = None,
-            value_dtypes: list[numpy.dtype[Any]] | None = None,
+            target_kernels: Sequence[ScalarKernel],
+            source_kernels: Sequence[ScalarKernel],
+            strength_usage: Sequence[int] | None = None,
+            value_dtypes: Sequence[DTypeLike] | DTypeLike | None = None,
```
alexfikl, line 310: These seemed ok to loosen a bit?

sumpy/tools.py
```diff
@@ -333,21 +343,21 @@ def __init__(self,
...
+    def nresults(self) -> int:
         return len(self.target_kernels)
 
     @property
     @abstractmethod
     def default_name(self) -> str:
         pass
 
-    def get_kernel_scaling_assignments(self):
+    def get_kernel_scaling_assignments(self) -> Sequence[lp.Assignment]:
```
alexfikl, line 369: This caused a few new errors in e.g. `p2p.py` because we're doing `self.get_kernel_scaling_asssignments() + ["for itgt, isrc"]` and addition won't work in general.

sumpy/tools.py
```diff
@@ -717,61 +733,85 @@ def to_complex_dtype(dtype: DTypeLike) -> np.dtype[np.complexfloating]:
...
+    """Workalike for :class:`pyopencl.ProfilingInfoGetter`."""
     start: int
     end: int
 
 
-def get_native_event(evt):
-    from pyopencl import Event
-    return evt if isinstance(evt, Event) else evt.native_event
+@dataclass(frozen=True)
+class ProfilingEvent:
```
alexfikl, line 758: These profiling events got reworked a bit, although they're not used much anymore (from the `arraycontext` port, if I recall correctly?).

sumpy/tools.py
```diff
@@ -1086,19 +1133,18 @@ def run_opencl_fft(
...
 
 
-def __getattr__(name: str):
-    replacement_and_obj = _depr_name_to_replacement_and_obj.get(name)
-    if replacement_and_obj is not None:
-        replacement, obj, year = replacement_and_obj
+def __getattr__(name: str) -> Any:
+    replacement_and_year = _depr_name_to_replacement_and_obj.get(name)
+    if replacement_and_year is not None:
+        obj, year = replacement_and_year
```
alexfikl: This seems to have been broken? I guess nobody was using it. Can we delete it? :grin: It says "2023"...

sumpy/fmm.py
```diff
@@ -143,51 +156,53 @@ def __init__(self,
...
 
     @memoize_method
-    def get_base_kernel(self):
+    def get_base_kernel(self) -> ScalarKernel:
         from pytools import single_valued
         return single_valued(k.get_base_kernel() for k in self.target_kernels)
 
     @memoize_method
-    def multipole_expansion(self, order: int):
+    def multipole_expansion(self, order: int) -> MultipoleExpansionBase:
```
alexfikl, line 180: I'm not sure about returning the base classes here. It makes sense, but they're currently not implemented very nicely for something like this (e.g. `pyright` complains `MultipoleExpansionBase` is not callable because it has no `__call__` stub).

sumpy/tools.py
```diff
@@ -1022,11 +1064,11 @@ def get_opencl_fft_app(
 
 def run_opencl_fft(
         actx: ArrayContext,
-        fft_app: tuple[Any, FFTBackend],
-        input_vec: Any,
+        fft_app: tuple[lp.TranslationUnit | VkFFTApp, FFTBackend],
```
alexfikl: Not a big fan of this being a tuple (the first and second element could get out of sync?). Any reason to not just do an `isinstance`?

sumpy/fmm.py
```diff
@@ -538,34 +587,32 @@ def output_zeros(self,
...
-
-        assert (
-                isinstance(potentials, np.ndarray)
-                and potentials.dtype.char == "O")
-
-        def reorder(x):
+    @override
+    def reorder_potentials(
+            self, potentials: Array | obj_array.ObjectArray1D[Array]
+        ) -> Array | obj_array.ObjectArray1D[Array]:
```
alexfikl: This should also be added to `boxtree`. That currently just takes `Array`.

sumpy/fmm.py
```diff
@@ -919,16 +988,20 @@ def multipole_to_local(self,
...
         return local_exps
 
-    def eval_multipoles(self,
+    @override
+    def eval_multipoles(
+            self,
             actx: ArrayContext,
-            target_boxes_by_source_level, source_boxes_by_level, mpole_exps):
+            target_boxes_by_source_level: obj_array.ObjectArray1D[Array],
+            from_sep_smaller_by_level: obj_array.ObjectArray1D[BuiltList],
```
alexfikl, line 1022: Not quite sure why these had such different argument names? Went with the version from `boxtree`.

sumpy/fmm.py
```diff
@@ -1082,15 +1167,19 @@ def eval_locals(self,
 
         return pot
 
-    def finalize_potentials(self, actx: ArrayContext, potentials):
+    @override
+    def finalize_potentials(self, actx: ArrayContext, potentials: Array) -> Array:
```
alexfikl: This should probably also be a `Array | ObjectArray`, but needs a little update in `boxtree` as well.

sumpy/tools.py
inducer: Oh sure! Just make sure it's in whatever uv uses for dev dependencies, so that it's not pulled in at runtime.

sumpy/tools.py
inducer: Totally fine.

sumpy/__init__.py
inducer: Looks like it!

sumpy/tools.py
inducer: I agree, that would be nicer.

sumpy/tools.py
inducer: So be it. Arithmetic in general is a bit of a mess.

sumpy/tools.py
inducer: "Keeping it vaguely functional" is a good bar to use.

sumpy/tools.py
inducer: Yeah, kill it.

inducer:
Thanks for sifting through this!

sumpy/tools.py
alexfikl: Killed: [8f35910](https://github.com/inducer/sumpy/pull/300/commits/8f359103bd25848aca93f454742fe7c732b6d927)

sumpy/tools.py
alexfikl: [be09448](https://github.com/inducer/sumpy/pull/300/commits/be0944891295613d75e97e0dafe527a463252b3f) refactors this a bit.

sumpy/test/test_fmm.py
```diff
@@ -128,58 +133,66 @@ def test_sumpy_fmm(
...
-    if fft_backend == "pyvkfft":
+    if fft_backend == FFTBackend.PYVKFFT and isinstance(actx, PyOpenCLArrayContext):
         from pyopencl.characterize import get_pocl_version
-        if (isinstance(actx, PyOpenCLArrayContext)
-                    and get_pocl_version(actx.queue.device.platform) >= (7,)):
+        pocl_version = get_pocl_version(actx.queue.device.platform)
+
+        if pocl_version and pocl_version > (7,):
             pytest.skip("pocl 7 and pyvkfft don't get along: "
                         "https://github.com/pocl/pocl/issues/2069")
```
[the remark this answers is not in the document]
alexfikl: Oops. Fixed.

sumpy/tools.py
```diff
@@ -315,7 +334,7 @@ def __init__(self,
 
         if not isinstance(value_dtypes, Sequence):
             value_dtypes = [np.dtype(value_dtypes)] * len(target_kernels)
-        value_dtypes = [np.dtype(vd) for vd in value_dtypes]
+        value_dtypes = tuple(np.dtype(vd) for vd in value_dtypes)
```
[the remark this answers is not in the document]
alexfikl: Fixed.

alexfikl:
@inducer This should be good for a look too, whenever you have the time.

inducer:
Thx!

PR #303, opened 2026-09-04
Add SympyToPymbolicMapperWithSymbols

alexfikl:
This is needed in #279 (mostly to move `pi` back from pymbolic).

`PymbolicToSympyMapperWithSymbols` also implements a special `map_subscript`, but that can't be unambiguously roundtripped, so I left it alone.

### DIFF
--- a/sumpy/symbolic.py
+++ b/sumpy/symbolic.py
@@ -95,6 +95,9 @@ _find_symbolic_backend()
 
 # }}}
 
+
+# {{{ symbolic expressions
+
 if TYPE_CHECKING or not USE_SYMENGINE:
     import sympy as sym
 
@@ -169,6 +172,8 @@ else:
 
     def unevaluated_pow(a: Expr, b: complex | Expr) -> Expr:
         return Pow(a, b, evaluate=False)
+
+# }}}
 
 
 # {{{ debugging of sympy CSE via Maxima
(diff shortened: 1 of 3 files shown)


sumpy/symbolic.py
```diff
@@ -364,6 +375,45 @@ def map_call(self, expr: prim.Call) -> sym.Basic:
         return PymbolicToSympyMapper.map_call(self, expr)
 
 
+class SympyToPymbolicMapperWithSymbols(SympyToPymbolicMapper):
+    if USE_SYMENGINE:
+        @override
+        def map_Constant(self, expr: object) -> Expression:
```
alexfikl, line 381: Not a big fan of that `if`, but at this point the inheritance of these things is already split based on `USE_SYMENGINE`, so not quite sure there's a cleaner way to do it..?

inducer: I can't think of something nicer. Seems fine to me.

inducer:
Thx!

isuruf (commented):
We can rename `Constant` to `NumberSymbol` in symengine.py to match sympy. Then keep Constant as an alias.

alexfikl:
> We can rename `Constant` to `NumberSymbol` in symengine.py to match sympy. Then keep Constant as an alias.

I don't feel too strongly about that. It would be nice, for sure, but there will always be some mismatch between the two. Are you making a release for symengine 0.15? :grin:

isuruf:
> Are you making a release for symengine 0.15? 😁

Yes, probably this weekend.

PR #305, opened 2026-09-04
Rework sympy to pymbolic mappers

alexfikl:
The mappers previously inherited from either the Sympy or SymEngine variants based on `USE_SYMENGINE`. This meant that you couldn't use them at all for the other one, e.g. if `USE_SYMENGINE=True` there was no way to convert `sympy` expressions.

This implements both variants (with some code duplication) and adds `to_pymbolic` and `to_symbolic` helper functions that do the actual dispatch based on `USE_SYMENGINE`.

### DIFF
--- a/sumpy/codegen.py
+++ b/sumpy/codegen.py
@@ -64,11 +64,9 @@ logger = logging.getLogger(__name__)
 
 
 __doc__ = """
-
 Conversion of :mod:`sympy` expressions to :mod:`loopy`
 ------------------------------------------------------
 
-.. autoclass:: SympyToPymbolicMapper
 .. autofunction:: to_loopy_insns
 """
 
@@ -76,26 +74,6 @@ Conversion of :mod:`sympy` expressions to :mod:`loopy`
 def wrap_in_cse(expr: Expression,
                 prefix: str | None = None) -> prim.CommonSubexpression:
     return prim.make_common_subexpression(expr, prefix, wrap_vars=False)
-
-
-# {{{ sympy -> pymbolic mapper
-
-_SPECIAL_FUNCTION_NAMES = frozenset(dir(sym.functions))
-
-
-class SympyToPymbolicMapper(sym.SympyToPymbolicMapper):
-    @override
-    def not_supported(self, expr: object) -> Expression:
-        if isinstance(expr, int):
-            return expr
-        elif getattr(expr, "is_Function", False):
-            func_name = sym.SympyToPymbolicMapper.function_name(self, expr)
-            return prim.Variable(func_name)(
-                    *tuple(self.rec(arg) for arg in expr.args))
-        else:
-            return sym.SympyToPymbolicMapper.not_supported(self, expr)
-
-# }}}
 
 
 # {{{ bessel -> loopy codegen
@@ -758,8 +736,7 @@ def to_loopy_insns(
     assignments = list(assignments)
 
     # convert from sympy
-    sympy_conv = SympyToPymbolicMapper()
-    pymbolic_assignments = [(name, sympy_conv(expr)) for name, expr in assignments]
+    pymbolic_assignments = [(name, sym.to_pymbolic(expr)) for name, expr in assignments]
 
     flat = FlattenMapper()
     bdr = BesselDerivativeReplacer()
--- a/sumpy/symbolic.py
+++ b/sumpy/symbolic.py
@@ -39,7 +39,7 @@ __doc__ = """
 
 import logging
 import math
-from typing import TYPE_CHECKING, ClassVar, cast
+from typing import TYPE_CHECKING, Any, ClassVar, cast
 
 from typing_extensions import override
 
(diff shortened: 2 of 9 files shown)


alexfikl:
Hmm.. this needs a bit more work. It needs to make a distinction between `USE_SYMENGINE` and `HAS_SYMENGINE`..

sumpy/codegen.py
```diff
@@ -78,26 +76,6 @@ def wrap_in_cse(expr: Expression,
...
-
-class SympyToPymbolicMapper(sym.SympyToPymbolicMapper):
-    @override
-    def not_supported(self, expr: object) -> Expression:
-        if isinstance(expr, int):
-            return expr
-        elif getattr(expr, "is_Function", False):
-            func_name = sym.SympyToPymbolicMapper.function_name(self, expr)
-            return prim.Variable(func_name)(
-                    *tuple(self.rec(arg) for arg in expr.args))
```
alexfikl, line 94: This seems to be handled in the base class now exactly the same, so I removed this whole thing.

sumpy/symbolic.py
```diff
@@ -404,97 +409,97 @@ def from_sympy(cls, expr: Symbol) -> SpatialConstant:
...
-        def map_NumberSymbol(self, expr: sym.NumberSymbol) -> Expression:
-            if expr is pi:
-                return prim.Variable("pi")
-            elif expr is I:
-                return prim.Variable("I")
+class PymbolicToSympyMapper(PymbolicToSympyLikeMixin, PymbolicToSympyMapperBase):
+    symbols: bool
+
+    def __init__(self, *, symbols: bool = False) -> None:
+        self.symbols = symbols
```
alexfikl, line 456: This also adds a little attribute to the mapper instead of having another mapper altogether. Good idea? Seemed cleaner..

sumpy/symbolic.py
```diff
@@ -506,6 +511,107 @@ def not_supported(self, expr: object) -> Expression:
...
+            return prim.Variable("pi")
+        else:
+            return super().map_NumberSymbol(expr)
+
+    @override
+    def map_ImaginaryUnit(self, expr: sym.ImaginaryUnit) -> Expression:
+        if not self.symbols:
+            return super().map_ImaginaryUnit(expr)
+
+        return prim.Variable("I")
```
alexfikl, line 568: I don't think this is used anywhere (will be in #279), but it didn't roundtrip `I` correctly at all before.

alexfikl (commented):
Worth noting: this is pretty breaking because `PymbolicToSympy` just does `sympy` now. From what I can tell, these are undocumented classes that are not used by `pytential` or anyone.. so not a big deal? :grin:

sumpy/symbolic.py
```diff
@@ -506,6 +511,107 @@ def not_supported(self, expr: object) -> Expression:
...
+
+        @override
+        def map_Constant(self, expr: object) -> Expression:
+            if not self.symbols:
+                return super().map_Constant(expr)
+
+            if expr is self.sym.pi:
+                return prim.Variable("pi")
+            else:
+                return super().map_Constant(expr)
```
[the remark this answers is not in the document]
alexfikl: This is wrong. The `I` never gets here because of `map_ComplexDouble` or whatever.

sumpy/symbolic.py
inducer: It's not well-explained. I have a vague sense of what it does, but that's based on Ctrl+F. Add a docstring?

inducer, line 456: Shouldn't it be declared in the mixin instead?

inducer, line 456: Is `symbols=False` ever needed really?

sumpy/symbolic.py
inducer: Shouldn't this just become `1j`?

inducer (commented):
Thanks! A few comments/questions below.

sumpy/symbolic.py
alexfikl: I'm not sure? This tries to roundtrip with `SympyToPymbolic` mostly + places like `HelmholtzKernel` use `var("I")`..

sumpy/symbolic.py
alexfikl: Sure. I didn't put it in there because in my head mixins don't carry state, but I guess it doesn't matter that much. `pyright` is also very upset by the mixin :cry:

alexfikl: I think at least some of the codegen needs it so that things like `var("pi")` from the kernels get transformed into numbers (which is what happens with `symbols=False`). Not sure this is the best place to do that..

inducer: Well, if methods in the mixin rely on a bit of data, the same class should declare what it's supposed to be. (IMO)

alexfikl: Moved in https://github.com/inducer/sumpy/compare/f4c6f95906def291811757bc6aa8e54fb6d09fff..eee22cd78a2bec98513172b412185a449811bd97.

alexfikl: Added some docs in https://github.com/inducer/sumpy/compare/f4c6f95906def291811757bc6aa8e54fb6d09fff..eee22cd78a2bec98513172b412185a449811bd97.

inducer:
Thx!

PR #306, opened 2026-09-11
Fix crashing Hankel1.n() calls

alexfikl:
From what I can tell, the test was hard segfaulting on `main` because it was getting somewhere into `symengine` that assumes function wrappers of this type always return numeric values on `.n()`. It was then trying to cast that to a number and crashing..

The fix here works if the arguments to `Hankel1`/`BesselJ` are actually numeric, by going through `sympy` to evaluate them. If they're not numeric it still crashes, so that needs some code in `symengine` and `symengine.py`.


inducer:
Thx!

PR #307, opened 2026-09-12
gpu-p2p small fixes

ShawnL00:
- Read the source dtype from the explicit sources parameter, which is not part of kwargs.
  - Read the strength dtype from the first strength array rather than from the containing sequence.
  - Keep one source-axis entry after concatenating the two local arrays to ensure the strict zip has matching lengths.


inducer:
Thx!

